PR #276, opened 2022-10-08
Make board_approval_date nullable on StockClass.schema.json

### LINKED ISSUE #251: [Enhancement] - Make `board_approval_date` nullable on StockClass.schema.json
Right now, this is a required value. This is problematic in circumstances where the information might not yet be available / exist.

As part of the conversation around this, we kicked around the idea of whether we might want two data types similar to PreReleaseOmission:

1. InformationNotYetAvailable (or something like that)
2. Redacted - where you might want to share OCF data but want to omit certain things. 

### Discussed in https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/discussions/238

<sup>Originally posted by **jacobyavis** June 28, 2022</sup>
What if the stock class is not yet approved, or a system does not have this data?
If it's a required field I can see someone coding in the filing date as the approval date for example, which isn't necessarily accurate.</div>

xudiera:
#### What type of PR is this?

/kind enhancement

Makes `board_approval_date` nullable on `StockClass`. See #238 for why.

Fixes #251


JSv4 (approved):
LGTM! Thanks for the contribution :-)

PR #285, opened 2022-10-17
Fixes Windows compatibility for docs generator

### LINKED ISSUE #186: [Bug]: Windows Compatiblity for Doc Generator
### Current Behavior:

If you run `npm run generate-docs` on a windows machine (or at least on mine), SchemaReader.ts throws an error that appears to be a windows-specific inability to support our `$ref` urls ? 

Doc Generator Error

I vaguely recall having a similar issue when I built the python validator way back when. Not sure what the root issue is here. I established it was def due to the OS though as I manually disabled the husky pre-commit check, pushed the changes to my branch, cloned the branch on my Linux box, and then generated the docs with no issues. 

### Expected Behavior:

Doc generation should work the same on Windows, Max and Linux.

### Steps to Reproduce:

Run the doc generator in a Windows environment (why would you, I know, but someone is bound to do it, like me 😆 )

### Environment:

- OS: Windows 10
- Node: 14.17.6
- npm: 6.14.15

Nope

xudiera:
#### What type of PR is this?

/kind bug

Running `npm run generate-docs` on Windows fails due invalid URL scheme for Windows' files.

Fixes #186

### DIFF
--- a/utils/generate-docs/Schema/ExamplesReader.ts
+++ b/utils/generate-docs/Schema/ExamplesReader.ts
@@ -2,6 +2,8 @@ import readdirp, { EntryInfo } from "readdirp";
 
 import { ExampleJson } from "./Examples.js";
 
+import { pathToFileURL } from "url";
+
 /**
  *  Reader parses a file tree of example implementations of JSON data conforming
  * to the OCF and loads into JS objects.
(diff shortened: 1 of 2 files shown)


xudiera:
@JSv4, the error you are seeing is related to Node v18.4.0. I am using Node v16.13.1 which is the one specified on `.node-version` and everything works fine on it. See below:
image

JSv4 (approved):

PR #293, opened 2022-10-19
Naveed/issue 281 render docs on gh pages

### LINKED ISSUE #281: [Enhancement]: Suggest Using MkDocs or TypeScript Equivalent to Render Our Docs into a GitHub Page

I've recently started using [MkDocs](https://github.com/mkdocs/mkdocs) to render project documentation for some other projects I'm working on. It's a powerful but still light-weight solution that takes valid Markdown and converts it into very nice html, complete with nav and a much nicer look and feel. I think the [Python Hatch Project Docs](https://hatch.pypa.io/latest/) provide a great look at what's possible with the tool. 

The best part is you can [deploy the resulting html page instantly to GitHub pages](https://www.mkdocs.org/user-guide/deploying-your-docs/). I believe you could do all of this with an action. 

Obviously this would introduce some further deployment complexity (and Python dependencies), but I think we should consider this (or something like it) to further package and serve the docs in the near future. If anyone is aware of something similar written in TypeScript, I'm all ears. 

### Why is this Needed?

Our Markdown docs are fantastic from a content perspective. They have a lot of content and are comprehensive. I think they're getting a bit unwieldy, however, and GitHub markdown isn't the prettiest. Searching them is hard. Browsing the README can be overwhelming, I'd imagine. 

Using a tool like this lets us leverage our existing Markdown docs and toolchain while adding substantial improvements in usability. 

Check out the [MkDocs project page](https://github.com/mkdocs/mkdocs) and the example docs from the [Hatch project](https://hatch.pypa.io/latest/).

naveedn:
#### What type of PR is this?
/kind documentation

- Draft PR to create static site documentation with Mkdocs that is deployed to github pages to allow users to easily browse our documentation. 
- This is a DRAFT PR because this has been untested, also there are some wonky parts to the static site that I could use some help iterating on! 

Demo Site: https://naveedn.github.io/Open-Cap-Format-OCF/

Fixes #281


JSv4:
Sweet @naveedn! This looks awesome. Would love to discuss this with folks on our next call. Yeah, MkDocs is such a great fit for our toolchain. It makes it so easy to add much-needed nav on top of our docs without needing to make many (any?) changes to the doc generation utilities we have. The only thing that we should probably talk through is whether we're ok adding Python dependencies. 

I'm also heavily using MkDocs for another project I'm working on in my spare time, so I'm pretty familiar with it and very happy to tag-team and iterate with you. I've borrowed a lot of my configs from the [Hatch Project](https://hatch.pypa.io/latest/)

naveedn:
Updated the styles, favicon, and navigation layouts, pushed them to https://naveedn.github.io/Open-Cap-Format-OCF/

**TODO:**
- [ ] Update Readme with python install instructions for mkdocs
- [ ] Fix the Github check (or possibly combine the two similar generate docs jobs)
- [ ] Determine what we want to do for Readme (i had to make modifications to make images render)
- [ ] Determine if we want the static site docs merged in the repo or ignored? (leaning towards keeping them in)

naveedn:
Was able to get the github action to work correctly on my fork, but I am getting a 403 on the main OCF repo. I think I'll need help getting it to work properly. 

I'll work next on breaking up the main readme file into separate files / sections, while I wait for the meeting to discuss the github action. 

In addition, in my browsing, I came across https://github.com/galamdring/mkdocs-json-schema-plugin, which we could optionally leverage in the future if we don't want to custom generate the jsonschema into markdown files.

JSv4 (changes requested):
@naveedn, great work again on this. Couple thoughts on how to solve the two problems remaining to deploy. 

First, to authorize the deployment, the problem appears to be that __within the action container__ the publish command is not authorized to push the docs as it doesn't know how to access the `GITHUB_ACTION` credentials GitHub injects into the action containers. Check out how this third-party action does it: 

https://github.com/mhausenblas/mkdocs-deploy-gh-pages/blob/19d5a44e72e4d7cfc7792eb67b95c15edb64d774/action.sh#L35-L44

Second, let's try to get the nav links to work before pushing the docs to the outside world. I think we can achieve this using these steps:

1. In your workflow, we want to regenerate the docs __before__ generating the site. We need a command line flag in the doc generator to change the root url. That should be pretty easy. 
2. Once you re-generate the docs, make the site using mkdocs and then publish using the approach outlined in the third-party action above (feel free to use the action too). 
3. Just be super careful that the changed docs are not committed to the repo too. If you want to commit the site subdir, be careful you only grab that folder.

Make sense?

JSv4:
> @naveedn, great work again on this. Couple thoughts on how to solve the two problems remaining to deploy.
> 
> First, to authorize the deployment, the problem appears to be that **within the action container** the publish command is not authorized to push the docs as it doesn't know how to access the `GITHUB_ACTION` credentials GitHub injects into the action containers. Check out how this third-party action does it:
> 
> https://github.com/mhausenblas/mkdocs-deploy-gh-pages/blob/19d5a44e72e4d7cfc7792eb67b95c15edb64d774/action.sh#L35-L44
> 
> Second, let's try to get the nav links to work before pushing the docs to the outside world. I think we can achieve this using these steps:
> 
>     1. In your workflow, we want to regenerate the docs **before** generating the site. We need a command line flag in the doc generator to change the root url. That should be pretty easy.
> 
>     2. Once you re-generate the docs, make the site using mkdocs and then publish using the approach outlined in the third-party action above (feel free to use the action too).
> 
>     3. Just be super careful that the changed docs are not committed to the repo too. If you want to commit the site subdir, be careful you only grab that folder.
> 
> 
> Make sense?

@robwise, any pointers on how to make the doc generator url root modular? I can go dig into the source too, just not sure when I'll have a minute to do it.

naveedn:
Hey! So here are some of my findings so far (before starting to code):

- Good catch on the third party deploy action, i'll give it a shot, see if I can get it to work
- secondly, the static site will always run after the docs are generated, because of the `needs` param [here](https://github.com/naveedn/Open-Cap-Format-OCF/blob/naveed/issue-281-render-docs-on-gh-pages/.github/workflows/check-gen-docs.yml#L44). Screenshot included.
- Reading through the schema generator, it seems that it is URL unaware. All of the nav links are defined in the [schema files themselves](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/main/schema/enums/ConversionRightType.schema.json#L3) using [jsonschema base uri rules](https://json-schema.org/understanding-json-schema/structuring.html#base-uri). In this case, it might just be a simple search and replace in the IDE vs having to write any code -- but I want to double check that is our intent. If so, I think it might be cleaner / easier to do this in a separate PR, what do you think?

naveedn:
Screen Shot 2022-11-06 at 4 58 27 PM

Tried using third party, still got the 403 error. I am confident it is not a code error because it succeeded on my forked version as well. I believe the correct solution here is for you to merge the PR, because **pull-requests from forked repositories will be automatically given read only scopes** (source: https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/) and will therefore continue to fail. Another option to explore is PAT (Personal Access Tokens), but I'm less sure of that.

naveedn:
Other Notes / TODOs:
- [ ] SchemaGenerator code needs to point at docs/Readme for the find-and-replace logic for [TableOfContents](https://github.com/naveedn/Open-Cap-Format-OCF/blob/naveed/issue-281-render-docs-on-gh-pages/utils/generate-docs/Schema/TableOfContents.ts#L19)
- [ ] Need to add python dev environment instructions.

JSv4:
Hey @naveedn, thanks for the further investigation. Yeah, I think any changes to the doc generator should be a separate PR. I think it'll be cleaner letting us change the root as a) we'll probably want to be able to do that for specific releases anyway and b) there are a lot of links and find & replace in bulk always makes me shudder. You always think it can't go wrong... but I've been burned too many times ;-). 

I can work on a separate PR and then we can rebase this. After that, happy to merge this and hopefully that'll resolve the permission issues.

JSv4:
@naveedn, check out #314 when you get a sec. Maybe give the outputs from the generator a test on your page before we merge this?

JSv4:
Okay, I think I finallllyyy figured this out. MkDocs _should_ be able to convert valid links such that the markdown in our repo is automatically browsable in MkDocs. As we've seen, however, this is NOT working. 

The reason is MkDocs only supports relative links (e.h. `/docs/schemas/files/ManifestFile.md` is INVALID but, from a README.md in the /docs folder, `schemas/files/ManifestFile.md` is valid). This kinda sucks though because our markdown files are nested in places and we use absolute paths in some places in the doc generator template - e.g. the "Documentation Home" link is added to EVERY page as `/README.md`. Also, having the README.md in the repo root and not in/docs adds another layer of complexity to this. 

My suggest here is we:

1. Modify the doc generator to calculate relative paths.
2. Move the README to /docs and delete README in repo root.
3. The one exception is the jsons. We need to construct absolute urls for those for now and then find a way to render them into markdown and serve them in our MkDocs pages. There is an extension that seems like it might work but I've only gotten it working halfway.

JSv4:
Yup, I was right. Got MkDocs to automatically convert the links in this form (example for `StockLegendTemplatesFile.md`):

```code
:house: [Documentation Home](../../README.md)

---

### File - Stock Legend Templates

`https://opencaptablecoalition.com/schema/files/StockLegendTemplatesFile.schema.json`

**Description:** _JSON containing file type identifier and list of stock legend templates_

**Data Type:** `OCF_STOCK_LEGEND_TEMPLATES_FILE`

**Composed From:**

- [schema/primitives/files/File](../primitives/files/File.md)

**Properties:**

| Property  | Type                                                                                                                     | Description                               | Required   |
| --------- | ------------------------------------------------------------------------------------------------------------------------ | ----------------------------------------- | ---------- |
| file_type | **Constant:** `OCF_STOCK_LEGEND_TEMPLATES_FILE`</br>_Defined in [schema/enums/FileType](../enums/FileType.md)_ | Object type field                         | `REQUIRED` |
| items     | [ [schema/objects/StockLegendTemplate](../objects/StockLegendTemplate.md) ]                                    | List of OCF stock legend template objects | `REQUIRED` |

**Source Code:** [schema/files/StockLegendTemplatesFile](../../../schema/files//StockLegendTemplatesFile.schema.json)

Copyright © 2022 Open Cap Table Coalition.

```

JSv4:
@naveedn, rewriting the doc generator to generate relative links and then figure out a way to also package up the jsons took fooorrrreeeevvvveeer, but I did get it all working. Check it out - https://jsv4.github.io/OCF-Docs-Test/. 

I need to put together a PR, but here's the main things we have to change:

1) I had to restructure the docs folder a bit
2) I had to create a MkDocs action to copy schemas into /docs at doc build time
3) I had to move away from that third-party MkDocs action as it uses a container and pulls the docs from the repo. I need to do some manipulation to the docs folder (see 2 above) to get what we want that we don't want to commit to the repo itself. That means we need to a) generate new docs b) copy in necessary additional files and c) deploy docs in a single action. 
4) I modified the doc generator to create relative links. It was actually somewhat complex because any MD file can be arbitrarily nested in our repo and it can refer to any other md file with any other level of nesting. I basically had to pass paths from the doc generator object generating md for a given object to the doc generator object instance generating the links for the target schema.

naveedn:
Closing this as @JSv4 has added a lot of changes on top of the branch and we will merge that in instead.

PR #296, opened 2022-10-25
Add Transfer Documentation / How-To

JSv4:
#### What type of PR is this?

/kind documentation

#### What this PR does / why we need it:

Adds a detailed walkthrough of how to model a basic transfer in OCF. 

#### Which issue(s) this PR fixes:

Fixes #209

PR #303, opened 2022-11-08
[Enhancement] Add a StockClassConversionRatioAdjustment Transaction

### LINKED ISSUE #302: [Enhancement]: Add Transaction to Track Adjustments to Stock Class Conversion Ratios

When doing some work to generate a pro forma / summary cap table from OCF today, it occurred to me and my colleagues that we are missing some important event history in the current spec. The current [schema/types/conversion_mechanisms/RatioConversionMechanism](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/main/docs/schema/types/conversion_mechanisms/RatioConversionMechanism.md) describes how class X converts to class Y where there is no re-pricing. Where that ratio changes, however, such as where there's anti-dilution and the conversion price (and thus conversion ratio) is adjusted (or the conversion price / ratio is simply amended, which can happen), we lose the history of the repricings entirely. This is a big omission for cap table tie-outs / diligence.  

I propose we pursue a Stock-Class-level transaction event like we did for splits. I have an accompanying PR illustrating how this would work. 

### Why is this Needed?

In cases where the conversion price changes, the change in the conversion price / ratio is calculated according to whatever was agreed / negotiated, but the resulting conversion price post-adjustment must be recorded somewhere. With the current spec, if you had to change the conversion mechanism field on the Stock Class object, you'd lose all preceding conversion price history. This is a pretty significant loss, arguably an unacceptable loss of fidelity. I'd say this borders on a bug.

Some more [background](https://www.cooleygo.com/glossary/broad-based-weighted-average-anti-dilution-protection/) on anti-dilution for those who want a refresher.

JSv4:
#### What type of PR is this?

/kind feature

When doing some work to generate a pro forma / summary cap table from OCF today, it occurred to me and my colleagues that we are missing some important event history in the current spec. The current StockClassConversionMechanism works **_unless_** there's anti-dilution and the conversion price (and thus conversion ratio) is adjusted. In such cases, the change in the conversion price / ratio is calculated according to whatever was agreed in the financing for a particular class of stock, but the resulting conversion price post-adjustment must be recorded somewhere. With the current spec, if you had to change the conversion mechanism on the Stock Class object, you'd lose the conversion price history. This is a pretty significant loss, arguably an unacceptable loss of fidelity.

What I've proposed here is a mechanism similar to what we have for [Stock Class Splits](/schema/objects/transactions/split/StockClassSplit.schema.json) - a Stock _Class_ (as opposed to _Issuance_)-linked event that tracks the change in the conversion price and resulting conversion ratio. 

Fixes #302


JSv4:
Realized this branch also incorporates #296 too. Sorry :-/

JSv4:
@jacobyavis, thanks for the review. I'm looking at my commit history above and I think I fat-fingered something in VSCode when trying to rebase... Some weirdness going on above.

Maybe I'll just revert to the changes you OK'ed and then I'll rebase PR #306 on `main` once the changes here merged in.

JSv4:
@jacobyavis, actually I think I forgot to switch to main when creating this branch, so it included my how-to changes in #296. The changes in the latest commit to this branch show updates to reflect the new RatioConversionMechanism proposed in #306. I'll do better next time ;-).

jacobyavis:
All good @JSv4 !

PR #304, opened 2022-11-08
add new type InterestRate and allow specifying an array of these on the NoteConversionMechanism

### LINKED ISSUE #249: [Bug] - Add Interest Accrual End / Terminator for Convertible
We need a field to track the start and end period / time / date for a convertible. Most likely needs to go in [NoteConversionMechanism](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/v1.0.0-beta/docs/schema/types/conversion_mechanisms/NoteConversionMechanism.md)

### Discussed in https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/discussions/234

<sup>Originally posted by **magonjr** June 27, 2022</sup>
We propose adding a field called Interest Accrual End Date to denote when a convertible note ceases accruing interest since, occasionally, the date the interest stops accruing is not the same as the conversion date. We only saw a conversion date field in OCF, so wasn’t sure if such an edge case could be handled.</div>

### LINKED ISSUE #262: Add support for multiple interest rates to Convertible Notes
See below:

Also noted that there is a lack of a end date / maturity date in the existing mechanisms. This should be optional to support accrual "forever" or stopping at that date.

### Discussed in https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/discussions/245

<sup>Originally posted by **jacobyavis** June 28, 2022</sup>
In Shareworks you can store multiples interest rates along with their effective dates on a convertible note.
E.g.
1% interest for the first year
1.5% interest for the second year
2% interest after that

The OCF format does not support this as there's only the one field for interest rate.</div>

jacobyavis:
#### What type of PR is this?

/kind feature

Some convertible notes specify multiple different interest rates along with their effective dates. Previously, the schema only allowed for one interest rate on the conversion mechanism.

Futher, it was assumed that interest accrual commences on issuance, but now it can be specified exactly when interest accrual begins.

This PR also adds an `accrual_end_date` as requested in issue #249

One last change is that the interest rate value is now stored using the Percentage type, instead of Numeric. E.g. it will now use value "0.0125" for 1.25% instead of "1.25".

Fixes #262 
Fixes #249


JSv4 (approved):
LGTM. The rules you set out in the descriptions re: the start and end dates make sense to me, though they just underline the need for a second-layer of validation to ensure that rules like this are followed (or at least that values inconsistent with the rules cannot be considered _valid_).

PR #306, opened 2022-11-08
Added `conversion_price` field to RatioConversionMechanism.schema.json

### LINKED ISSUE #305: [Enhancement]: Add conversion_price field to

As part of the TWG discussion around the proposal to fix issue #302, we decided we wanted to track `conversion_price` throughout the history of a stock object. Right now, you can back that out from the `original_issue_price` and the conversion ratio at issue, but it's not tracked explicitly. Though it's somewhat duplicative, we decided to include both fields on the `RatioConversionMechanism` itself. Then, the `StockClassConversionRatioAdjustment` transaction object proposed in #302 can just hold a new `RatioConversionMechanism` type in a single field rather than having the `adjusted_conversion_price` field in addition to a `RatioConversionMechanism` field holding just a ratio.

### Why is this Needed?

We want to track conversion price history throughout life of a Stock Class.

#302 will need to be re-based once this is merged.

JSv4:
#### What type of PR is this?

/kind feature

As part of changes to track changes in conversion ratio / conversion_price of stock classes, we decided to made conversion_price a field of the RatioConversionMechanism. Then, the Conversion Ratio Adjustment proposed in #302 can just contain a new RatioConversionMechanism, rather than a `conversion_price` field and a separate `RatioConversionMechanism` type. 

Fixes #305

### DIFF
--- a/docs/schema/objects/StockClass.md
+++ b/docs/schema/objects/StockClass.md
@@ -77,6 +77,10 @@
       {
         "conversion_mechanism": {
           "type": "RATIO_CONVERSION",
+          "conversion_price": {
+            "amount": "1.00",
+            "currency": "USD"
+          },
           "ratio": {
             "numerator": "1",
             "denominator": "1"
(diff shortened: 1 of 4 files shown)


JSv4:
@jacobyavis, can you approve this when you get a sec? Need that to merge it.

jacobyavis (approved):
LGTM John!

PR #307, opened 2022-11-08
Set npm to ~8.19 in package engines

pjohnmeyer:
#### What type of PR is this?
Developer experience

Establishing the acceptable `npm` version in the `engines` block of the package file will ensure developers are using the same version of `npm`, avoiding unexpected changes or thrashing of the lockfile.

N/A, follow-up from 2022-11-08 TWG call.


JSv4 (approved):
LGTM

PR #309, opened 2022-11-14
Add Stock Plan-level Pool Adjustment Transaction

### LINKED ISSUE #308: [Enhancement]: Add Stock Option Plan Pool Change

This information is pretty significant for the creation of pro formas and "traditional" cap tables in Excel. Similar to issue #302, we are not tracking these pool changes as we don't track changes to the Stock Plan objects, we just output the most current state. Like the fix for #302 proposed in PR #303, the fix would be to create an object-level "transaction" for Stock Plan objs to change the pools.

### Why is this Needed?

It's no longer a big paradigm shift to add object-level events, and it's very important for legal users.

Nope

JSv4:
#### What type of PR is this?

/kind feature

As discussed in issue #308, tracking pool changes is significant for legal practitioners and is a needed feature in the Excel-based exporter they legal working group is developing. Since we already started adding object-level transactions for Splits and Conversion Price adjustments, I think this is also important enough to add and doesn't introduce design patterns we haven't already OK'ed.

Fixes #308


jacobyavis:
This is a great solution to allow us to track adjustments in the transaction stack. Great work John :)

jacobyavis (approved):

JSv4:
> This is a great solution to allow us to track adjustments in the transaction stack. Great work John :)

Thanks, Jake! Appreciate the quick review.

PR #310, opened 2022-11-15
Remove `current_` prefix from the Stock Plan `current_shared_reserved` field

### LINKED ISSUE #308: [Enhancement]: Add Stock Option Plan Pool Change

JSv4:
#### What type of PR is this?

/kind feature

Matching change to PR #309. In addition to adding the transaction to change the pool, we want to change the naming scheme we have for `current_shares_reserved` and edit the description so it doesn't explicitly say we don't track pool sizes. 

Fixes #308, along with #309


jacobyavis:
This follow #309 nicely, thanks for this!

jacobyavis (approved):

PR #314, opened 2022-11-22
Make Doc Generator Urls Modular For MkDocs

### LINKED ISSUE #281: [Enhancement]: Suggest Using MkDocs or TypeScript Equivalent to Render Our Docs into a GitHub Page

JSv4:
#### What type of PR is this?

/kind feature

As discussed in PR #293, we want to use MkDocs to host our generated .MD docs. The problem, however, is the links in those docs are all relative to the repo root and assume that you'll use GitHub to navigate them. These don't work in MkDocs as it expects absolute urls. 

**To enable the MkDocs deploy, this PR provides:**

1. A way to replace the url roots in the documentation, and, since the source schemas may or may not be hosted in the MkDocs platform, we need a way to specify separate url roots for the schema source links vs other markdown documentation. 
  
    1. I wrapped the doc generator in command line args package `yargs`. 
    2. Now you can use the `----repoUrlRoot` arg to specify the url the repo lives at. You can still generate relative paths too... just pass an empty string for this argument OR, if you want to target a specific version of the docs, you can specify a specific tag in the docs with something like this: `tree/v1.0.0-beta/docs`. 
    3. You can use the `--docUrlRoot` arg to specify the root url of the documentation. For GitHub links, use `/docs`, otherwise use an absolute url like `https://naveedn.github.io/Open-Cap-Format-OCF`.

4. Since MkDocs apparently doesn't add extensions to its md files in the urls, we need a way to optionally drop the .md extension

    1. Use the `--addFileExtension` flag to add `.md` to the markdown links. Required for GitHub. Incompatible with MkDocs. 

5. Since the home README file we want to refer to may be hosted on MkDocs or the repo and the url form is different in both cases, we want a way to specify the README url separately. You **can't** specify this as a relative path from the root url because GitHub and MkDocs handle the relative paths separately. GitHub drops our README into `/README.md` whereas MkDocs puts it at the root page with no extension at all - e.g. ` https://naveedn.github.io/Open-Cap-Format-OCF/`.

    1. Use the `--indexPath` argument to pass the absolute url for the README (or relative path like `/README.md` if you want relative links still).

I should mention, **I created two generate-docs scripts in the package.json for our different deploy cases**. `npm run generate-docs` should generate exactly the docs we have, and these should go in the repo. `npm run generate-pretty-docs` should generate docs with links that target a GitHub page. Currently setup to target Naveed's page **but**, once we have things running, we can switch the url to our project GitHub page. 

Required for PR #293, and, with #293, fixes #281.


JSv4:
@naveedn, can you play around with the outputs from this in your GitHub page to see if the resulting markdown works as desired?

JSv4:
Per our TWG discussion this week (week of Nov 28th), I took a step back and tried to figure out why MkDocs wasn't just loading our docs properly. The reason was fairly simple - it requires relative links. The fix is fairly complex, however, as it means, for any given folder of md pages, they need to calculate different relative paths than other folders. I have version of the doc generator that can do this, but it's a very different approach from what I did here. I'm going to close this and open a new PR with my new code.

PR #319, opened 2022-11-25
Add VestingAcceleration transaction

### LINKED ISSUE #287: [Enhancement] - Create new vesting transaction to represent accelerated vesting
### Discussed in https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/discussions/271

<sup>Originally posted by **pjohnmeyer** August 31, 2022</sup>
## Why

Currently, there is no way to represent a vesting acceleration in the transaction log.

**Note:** This is not meant to replace milestone vesting via [VestingEvent](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/main/schema/objects/transactions/vesting/VestingEvent.schema.json). Those events reference a milestone that are part of the vesting terms, whereas acceleration is an "advance" on shares that are scheduled to be vested in the future.

## Proposal

Create a new event:

```json
{
  "object_type": "TX_VESTING_ACCELERATE", // required, const
  "reason_text": "some platform provided reason, not enumerated", // optional, string
  "quantity": "100", // required, numeric
  "date": "2022-02-02", // required, date
  ... remainder of Object / Transaction / SecurityTransaction fields
}
```

**Question** Do we need an optional "effective date" or "board approval date" on this as well?

### Do nothing
If we do not create a VestingAccelerate event, then any security that has vesting accelerated will need to create a new [VestingTerms](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/v1.0.0-b2/schema/objects/VestingTerms.schema.json) object unique to that security. While this is a valid alternative, acceleration is a common enough occurrence that I believe the schema should natively support it.

### Represent vesting acceleration as an event in the VestingTerms
We could add a new trigger and represent accelerated vesting as an event-based [VestingCondition](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/v1.0.0-b2/schema/types/vesting/VestingCondition.schema.json). Doing this, however, would greatly complicate the graph, and would require more parameterization of the condition. This has similar complexities expressed in #269 that I won't repeat here.
</div>

pjohnmeyer:
#### What type of PR is this?
/kind feature

See #287 

Fixes #287


jacobyavis:
Looks good to me, though I think it would be nice to include a sample of this transaction type. I believe we have examples of all other transaction types already.

jacobyavis (approved):

JSv4 (approved):

PR #330, opened 2022-12-15
Doc Generator Cleanup Fixes

### LINKED ISSUE #329: [Bug]: Doc Generator Cleanup - Revise Husky and Prettier Configs for New Doc Folder Structure
### Current Behavior:

Husky pre-commit hooks are failing because they're trying to add `/docs/schema` and `/README.md` to the staged files prior to commit. These paths are no longer valid. 

Also, prettier is running on the markdown, which I forgot we didn't want to do that. `.prettierignore` is pointed at the old `/docs/schema` dir and not `/docs/schema_markdown`.

### Expected Behavior:

Husky should work without issues and prettier should ignore generated markdown files.

### Steps to Reproduce:

Try to create a commit.

### Environment:

- OS: Ubuntu 22.04
- Node: 16.13.1
- npm: 8.1.2

PR already done.

JSv4:
#### What type of PR is this?

/kind bug

This does two things:

1. When I created the PR for the doc generator, I overlooked I needed to change some of the settings in the husky and prettier config. I fixed those config issues:
    1. The husky configs now point to `/docs/schema_markdown` instead of the old `/docs/schema`
    2. I've removed the pre-commit calls to git add `/README.md` as a) it moved and b) it is no longer automatically updated so there's no need for this call in a pre-commit hook.
    3. I forgot we were ignoring the markdown with prettier. I updated the `.prettierignore` to point at `/docs/schema_markdown` so we are no longer fixing the formatting of the markdown. This shows up here as a a large number of changed files as they've not been prettified whereas they were in the last merged commit / PR. 
2. In the process of creating the utility function to to build relative paths, I changed `schemaPathRelativeToSchemaDir()` in PathTools.ts to use the same `relativePathToOtherPath()` function that I created for the doc generator. That `relativePathToOtherPath()` function pulled code out of `schemaPathRelativeToSchemaDir()` BUT dropped the path separator at the start of the return string of the original code (presumably to deal with format MkDocs was expecting). This was causing generated copyright comments to have invalid urls that were missing one path separator. Since this change appears to only be affecting the copyright comment path generator, I am going to just add that slash back into the return statement of `schemaPathRelativeToSchemaDir()`.

Fixes #329


jacobyavis (approved):
Looks good to me!

PR #331, opened 2022-12-15
Add Pre and Post Flavors to SAFEConversionMechanism

### LINKED ISSUE #321: [Enhancement]: Add Pre/Post Money Safe Flag on SAFEConversionMechanism

Originally, before we came up with the conversion mechanism scheme, we had a pre/post money flag on the safe object. When we moved many of the SAFE conversion values into the SAFEConversionMechanism type, it looks like we inadvertently lost the pre/post flag:

https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/52e77094c64358c76b71e7fa9a3c994caf4be187/schema/types/conversion_mechanisms/SAFEConversionMechanism.schema.json

### Why is this Needed?

This is pretty commonly used and we ought to be able to track it.

Please confirm this flag wasn't moved somewhere else? We checked but please double-check.

JSv4:
#### What type of PR is this?

/kind feature

We currently can't track PRE-MONEY vs POST-MONEY SAFEs. We had planned to add this to the `SAFEConversionMechanism`, but this change was never committed for some reason. Opened a PR to add an enumeration for pre vs post (`ConversionTimingType`) and add a field `conversion_flavor` to the `SAFEConversionMechanism` to track this for any given SAFE (as the mechanism is part of the issuance). Added a `capitalization_definition` field to `SAFEConversionMechanism` to reflect what we did for Notes.

Also added an example showing this conversion mechanism type.

Includes changes from PR #330 

Fixes #321


jacobyavis (approved):
Thanks for bringing this back, looks good to me

schema/types/conversion_mechanisms/SAFEConversionMechanism.schema.json
```diff
@@ -28,9 +28,17 @@
     "conversion_mfn": {
       "description": "Is this an MFN flavored SAFE?",
       "type": "boolean"
+    },
+    "conversion_flavor": {
```
pjohnmeyer, line 32: @JSv4 Flavor seems an odd choice of words here. Feels too much like idiomatic US English to me. Perhaps we just call it `conversion_timing` to mirror the type?

JSv4: Haha, fair. I went back and forth on that. Give me a sec.

JSv4: Revised, @pjohnmeyer

PR #332, opened 2022-12-15
Remove PreReleaseOmission

### LINKED ISSUE #323: [Enhancement]: Make PreReleaseOmission Fields Simply Required Fields

As we developed the beta, we created a field for `PreReleaseOmission` for fields that we felt should be required but which there wasn't full support for amongst our vendors. Now that we're planning to introduce a stable release, we need a strategy to deal with this. My suggestion is to change the type's name from `PreReleaseOmisison` to `IntentionallyOmitted` or `NotSupported`. We also should discuss how we'd like to message the timeline for when `IntentionallyOmitted` or `NotSupported` type will be dropped for a given field - like a reverse deprecation. 

### Why is this Needed?

Once we release 1.0.0, we are no longer "pre-release", so, at the very least, this name needs to change. We also should take this opportunity to think about how to handle fields that we want to push the community to support in the near future - a sort of reverse deprecation. 

Nope

JSv4:
#### What type of PR is this?

/kind cleanup

As discussed in issue #323 and the TWG call on 12/13/22, this field type makes no sense after a stable release. We discussed alternatives, but ultimately decided, since this is only used in two places, two remove the concept entirely. As is always the case with OCF, we rely on exporting systems to provide accurate info or, where that's not possible, the best information they have. If they need to confirm this info on export or come up with a way to determine the appropriate values internally, they'll need to do that. 

Assumes merge of #330 (that PR affects a lot of generated files).

Fixes #323


schema/objects/transactions/issuance/StockIssuance.schema.json
```diff
@@ -40,15 +40,7 @@
     },
     "cost_basis": {
```
jacobyavis, line 41: I'm thinking cost_basis should be nullable. I think we all agreed on votes per share being required, but not cost basis.
If we follow the discussion link https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/discussions/218 the concern was around needing the field at all, and was agreed that having the field was necessary for handling odd cases where cost_basis is not/no longer calculable as quantity * share_price.

But I wouldn't say that's enough to make it a required field. If it's null we could assume the cost basis is quantity * share_price

JSv4: That works for me. So you're thinking we'll have it be a required field with null or a value instead of an optional field?

jacobyavis: It should be a regular optional field, which means removing "cost_basis" from the required array down below

JSv4: Cool @jacobyavis. Easy fix. Give me a sec.

JSv4: Fix is in, @jacobyavis

jacobyavis (approved):
Thanks John!

PR #339, opened 2022-12-16
Add `ShareNumberRange` Type and `share_numbers_issued` field to `StockIssuance`

### LINKED ISSUE #316: [Enhancement]: Add "numbered shares" field to stock issuance object
### Question:

Does the OCTF have a standard for dealing with numbered shares?

In several jurisdictions including Abu Dhabi Global Market which is used for many MENA-based startup holding companies, the authorizing body requires tracking of numbered shares. In other words, shares are non-fungible. (e.g. If Shareholder A transfers 50,000 shares, they don't just transfer a block of 50,000 shares, they transfer shares numbered from 1,000,001 to 1,050,000)

One potential option that has been explored is using the custom-id field for tracking these numbers but that field is still valuable for share certificate tracking in various jurisdictions.

JSv4:
#### What type of PR is this?

/kind feature

As discussed in issue #316, there are situations where shares are not fungible and each share must be capable of being tracked individually (or at least we must be able to identify grouped blocks of share numbers - e.g. shares 250 - 500). This PR adds a `ShareNumberRange` type to track these share number ranges and a `share_numbers_issued` field to `StockIssuance` events to track, for any given stock issuance and its `security_id`, which share numbers are in that issuance. It is an optional field, however, as this feature is not required in all jurisdictions. 

Fixes #316


jacobyavis (approved):
Looks good to me

PR #340, opened 2022-12-16
Add `StockClassAuthorizedSharesAdjustment` Event

### LINKED ISSUE #336: [Enhancement]: Add Stock Class Authorized Shares Change Transaction

Currently, we have no transaction object to change the shares authorized for a stock class. You can change this in your respective system of record, but you'll only get the most recent value, not the history, when you export the OCF. Since the authorized share history is extremely important information, and we have a design pattern to add this pretty easily, I am going to add a transaction for this.

### Why is this Needed?

Without the shares authorized history, you can't do a proper cap table tie-out. 

Nope

JSv4:
#### What type of PR is this?

/kind feature

Prior to this PR, our OCF exports will **only** include the most recently number of authorized shares of a given stock class. This is a deal-breaker for cap table diligence and tie-outs. Without this information, you can't ensure that the number of shares authorized was sufficient at any given point in the cap table's history. Unless there's a really compelling reason, strongly suggest we add this feature prior to v1.0.0.

Fixes #336


jacobyavis (approved):
Nice work! This is great

JSv4:
> Nice work! This is great

Thanks, @jacobyavis. Appreciate the quick review too!

PR #341, opened 2022-12-16
Changed Jurisdiction Field in SecurityExemption to a Plain String

### LINKED ISSUE #315: [Enhancement]: Provide alternative to ISO 3166-2 for Jurisdiction Fields

Replace the concept of using ISO 3166-2 for the Country Subdivision Code type with an enumerated list that matches valid jurisdictions for countries outside of the United States.

### Why is this Needed?

Currently, it looks like the Country Subdivision Code is a string field that is meant to be validated against the ISO 3166-2 code. 

The ISO 3166-2 list doesn’t really work outside the U.S. because the code does not match the jurisdictions of many other countries. For example, the subdivision of the United Arab Emirates are the seven Emirates but that does not match the jurisdictions list for the country which has some 52 jurisdictions including Abu Dhabi Global Market which is used by many regional startups. The same issue existing with the United Kingdom where the most common jurisdiction is actually “England & Wales”. 

I have not been able to find a good source for a master list of global legal jurisdictions for company formations, but it must exist somewhere.

JSv4:
#### What type of PR is this?

/kind cleanup

As discussed in issue #315, we realized that ISO 3166-1 alpha2 does not cleanly represent real legal jurisdictions, particularly outside of the U.S. Since there is no known enumeration of valid legal jurisdictions, we are moving this to free text. I left in the description that implementers should to try to use ISO 3166-1 alpha2 where appropriate but otherwise we rely on implementers to choose an appropriate value / convention. 

I wasn't 100% sure whether to change `country_subdivision_of_formation` field on `Issuer` to free text. I believe we agreed **not** to, but @jacobyavis, @pjohnmeyer, is that what you remember as well?

Fixes #315

### DIFF
--- a/docs/schema_markdown/schema/types/SecurityExemption.md
+++ b/docs/schema_markdown/schema/types/SecurityExemption.md
@@ -12,10 +12,10 @@ _Type representation of a securities issuance exemption that includes an unstruc
 
 **Properties:**
 
-| Property     | Type                                        | Description                                                                 | Required   |
-| ------------ | ------------------------------------------- | --------------------------------------------------------------------------- | ---------- |
-| description  | `STRING`                                    | Description of an applicable security law exemption governing the issuance  | `REQUIRED` |
-| jurisdiction | [schema/types/CountryCode](/CountryCode.md) | Country code of the jurisdiction of the applicable law (ISO 3166-1 alpha-2) | `REQUIRED` |
+| Property     | Type     | Description                                                                                                                                                                                                                                                         | Required   |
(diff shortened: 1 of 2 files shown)


jacobyavis (approved):

jacobyavis:
> I wasn't 100% sure whether to change country_subdivision_of_formation field on Issuer to free text. I believe we agreed not to, but @jacobyavis, @pjohnmeyer, is that what you remember as well?

I think this is the only 'jurisdiction' field so we should be good just changing this one, and leaving all the other fields that use CountryCode / CountrySubdivisionCode

PR #343, opened 2022-12-24
Add Stock Class ID to Valuation object

Sumukh:
#### What type of PR is this?

/kind feature 

This adds a reference to a `StockClass` in the `Valuation` field. Valuations may refer to differ classes of Stock (whereas the current assumption is that it points to the Common share class) 

I didn't create an issue ahead of time (but can next time if that's useful)


JSv4 (changes requested):
Hey @Sumukh, this is a great catch. Thanks for the contribution. I think we should make the new `stock_class_id` field required though, otherwise it won't be clear what the default should be in some cases (e.g. with multiple classes of common). Mind making that change and also merging in the latest changes from /main?

Sumukh:
Sounds good. I agree with the idea of making it required, but I didn't want to introduce a breaking change. I'll make that change and tag you again when it's ready for review

JSv4:
> Sounds good. I agree with the idea of making it required, but I didn't want to introduce a breaking change. I'll make that change and tag you again when it's ready for review

Thanks and 100% on the breaking change. Since we're still in beta, though, we're more ok with breaking changes atm. Once we release v1.0.0, though, we'll be much more rigorous about not introducing breaking changes without versioning up appropriately.

JSv4 (approved):
Looks great. Thanks again for the thoughtful contribution, @Sumukh!

PR #344, opened 2022-12-24
Avoid running the publish workflow on new PRs 

### LINKED ISSUE #333: [Enhancement]: Change Doc Publish Action to Only Run on Merge to Main

Docs should be pushed to MkDocs / gh-pages branch ONLY upon successful merge. Currently, they are getting pushed every time there's a PR opened (though, thankfully, the docs are only actually pushed if an authorized user opens the PR, so PRs from random people wouldn't successfully deploy the docs, though the GitHub action will try).

### Why is this Needed?

This current behavior could introduce lots of problems as unapproved changes will have their docs pulled into the public documentation site.

There appear to be some solutions here: https://github.com/orgs/community/discussions/26724

Sumukh:
#### What type of PR is this?

/kind cleanup 

This ensures that PRs do not run the generate-and-publish-mkdocs action. Those actions would have failed anyway on PRs from non-contributors (See #343). 

You can verify this worked because the `generate-and-publish` actions should not run in this PR

Fixes #333

### DIFF
--- a/.github/workflows/generate-and-publish-mkdocs.yml
+++ b/.github/workflows/generate-and-publish-mkdocs.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
 
 jobs:
   create-mkdocs-docs:


Sumukh:
I think the `Check Example OCF Files / Check Example OCF Files` just needs a re-run (It got some 429 errors when getting set up)

JSv4:
Thanks for the contribution, @Sumukh. Yup, looks like the test just needed to be re-run. I'll go ahead and merge this and then take a look at your other PR. I like the suggestion!

JSv4 (approved):
LGTM. Thanks!

PR #345, opened 2022-12-28
Add Detailed Terminal Transaction Guides

### LINKED ISSUE #198: [Enhancement]: Update Documentation to Note Terminal Transactions

In documentation, note terminal transactions (transactions that extinguish the issuance). 

### Why is this Needed?

It'd be nice.

### Discussed in https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/discussions/192

<sup>Originally posted by **pjohnmeyer** May  5, 2022</sup>
**Examples:**
- [quantity](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/053783239fcf919963b049ba09af8fce31fd4d11/schema/objects/transactions/transfer/StockTransfer.schema.json#L26)
- [amount](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/053783239fcf919963b049ba09af8fce31fd4d11/schema/objects/transactions/transfer/ConvertibleTransfer.schema.json#L26)

All four *Transfer transactions use the phrase "Quantity of non-monetary security units cancelled" or "Amount of monetary value cancelled". My initial suspicion is that these were indavertently copy/pasted from the *Cancel transactions and this should say "transferred".

This got me thinking though, when a transfer occurs, essentially security A is cancelled and "results" in securities B, C...

Should the transaction log contain a transfer and a cancel transaction for security A in this case, or only a transfer?</div>

JSv4:
#### What type of PR is this?

/kind documentation

Adds walk-through of how to trace transactions and which transaction are "terminal" - meaning (at least for purposes of this PR) that none of the original securities issued to the stakeholder as part of the transaction in question remain issued and outstanding to the same stakeholder. 

Fixes #198


JSv4:
@pjohnmeyer, @jacobyavis, do you guys agree with the definition I came up with for a "terminal transaction"? I think we could also fairly say almost all of our transactions are "terminal" in the sense that they extinguish a security_id, but I thought it was more natural to say the terminal transaction is where none of the securities from the original issuance remain issued and outstanding to the original holder.

docs/explainers/TraversingTransactions.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Terminal Transactions
+
+A **terminal** transaction is one which none of the securities originally issued as part of an issuance event remain issued and oustanding to the same holder. In many cases, our OCF transaction events allow _part_ of an issuance to be extinguished - e.g. a transfer or repurchase in part - but some of the original issuance will remain outstanding, in which case this type of event is **not** a terminal transaction. For example, where 1,000 shares are issued to Bob and then Bob transfers 500 to Frank, leaving 500 of the original 1,000 issued to Bob. While this is not a "terminal transaction:, where an OCF transaction acts on an issuance only in part, you'll see an optional `balance_security_id` field on the operative transaction. If a `balance_security_id` field is present, it indicates a remainder from the original `security_id` (as illustrated in the preceding example with Bob and Frank) is still held by the original holder but now under a new `security_id`.
```
pjohnmeyer, line 18: I thought the `balance_security_id` was also required for partial transfers / repurchases? I know the schema doesn't reflect that (because we don't have a separate schema type for e.g. PartialTransfer) but I thought a partial would also be terminal and result in a new security?

pjohnmeyer, line 18: Your visual aid doesn't match the example; the transferred shares in the partial transfer case should be 500 but is 1000.

docs/explainers/TraversingTransactions.md
```diff
@@ -0,0 +1,42 @@
...
+
+Keep in mind, however, that, even where the event acts on an issuance only in-part and some of the original issuance remains outstanding in the name of the original holder, the OCF spec requires that you us a new issuance event for this remainder of the original issuance. This remainder will have its own `security_id` for the "remainder" that remains outstanding in the name of the original holder. At this point, the original `security_id` from the original issuance is no longer in use.
+
+![](../images/OCF%20Terminal%20Transaction.png)
```
pjohnmeyer, line 22: Your visual aid doesn't match the example; the transferred shares in the partial transfer case should be 500 but is 1000.

docs/explainers/TraversingTransactions.md
JSv4: Yeah, that's what I meant what I said that I could see how basically all transactions are terminal. Went back and forth on the right nomenclature.  Very easily convinced to take your approach :-). Let me revise.

pjohnmeyer (approved):

PR #348, opened 2022-12-29
Update OCF Versions for Stable Release

JSv4:
#### What type of PR is this?

/kind cleanup

We have an enumeration of OCF version types. It was missing our beta 2 and beta 3 releases. It also needs to be updated for the 1.0.0 release.

Did not open an issue for this.


pjohnmeyer (approved):
Good catch. It occurs to me that having this as an enum is kind of useless because a particular validation can only work against the current version, but it is what it is today so this is fine.

PR #358, opened 2023-01-12
Fix unit tests

### LINKED ISSUE #355: [Bug]: Investigate and Repair Possible Test Failure that Doesn't Trigger Failure in GitHub Action
### Current Behavior:

As reported by @xudiera on today's TWG call, his branch had a test that failed for his PR that he needed to fix. The test and failure were both unrelated to changes he made, suggesting there is a test that's been failing in our test suite elsewhere. All of the GitHub actions passed, however, so we need to investigate to make sure that a) a test isn't silently failing and b), if it is, why our corresponding Action is not failing. 

### Expected Behavior:

If a test fails, GitHub Actions running it should fail too. 

### Steps to Reproduce:

@xudiera, can you add additional info here as needed?

xudiera:
#### What type of PR is this?

/kind bug

Unit test are failing. This PR address the unit tests failing

Fixes #355

### DIFF
--- a/docs/schema_markdown/schema/enums/AccrualPeriodType.md
+++ b/docs/schema_markdown/schema/enums/AccrualPeriodType.md
@@ -12,4 +12,4 @@
 
 **Source Code:** [schema/enums/AccrualPeriodType](../../../../schema/enums/AccrualPeriodType.schema.json)
 
-Copyright © 2022 Open Cap Table Coalition.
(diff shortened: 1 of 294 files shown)


xudiera:
TODO:

- [ ] Find out why the Action is not failing.

pjohnmeyer (approved):

PR #364, opened 2023-01-24
update the readme to specify rules around unknown files

### LINKED ISSUE #360: [Enhancement]: Explicitly support extraneous files

The OCF zip should be able to contain other files, such as cert images, documents, etc. This issue and PR will introduce an explicit mechanism for this.

### Why is this Needed?

Being able to export and import a cap table losslessly means including all the vendor-specific enhancements and features.

dan-pulley:
Fixes #360

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,13 @@ freely available for anyone to use, though changes must be made through the Coal
 
 ## How Does It Work?
 
-### OCF File Types
+### OCF File Rules
 
-OCF is a multi-file format designed to make it easy to split, compress or stream company capitalization tables.
-A valid OCF cap table is made up of JSON objects that match the schemas available in our repo in
-the [Schemas folder](/schema).
+OCF defines several JSON schemas that comprehensively represent elements of a Cap Table. The manifest file is the root file of OCF, and **references other files by filename**. The manifest file itself, and all the files referenced by it, **must** contain JSON files compliant with one of the OCF schemas.
+
+Any **"unknown file"** not referenced by the manifest may contain anything and should generally be ignored. This flexibility allows a system to export data unknown to OCF such as certificate PNG or document DOCX files. To facilitate ineropablity in the OCF ecosystem, any program that modifies OCF files **must not** modify or delete unrecognized files.
+
+### OCF File Types
 
 OCF objects are grouped and stored in eight file types (defined in our [file schemas folder](/schema/files)).
 
@@ -39,11 +41,11 @@ There are currently 8 file types that make up a cap table:
 8. [Vesting Schedules File(s)](/schema/files/VestingSchedulesFile.schema.json) - One or more files storing vesting
    schedules used by the issuer.
 
-**At the moment, we recommend combining all of these files into a single compressed file with a \*.ocf.zip extension:**
+**We recommend combining all of these files into a single compressed file with a \*.ocf.zip extension:**
 
 ![](docs/images/OCF%20Container.png)
 
-We are working on sample tooling to interact with compressed \*.ocf.zip files.
+We are working on sample tooling to interact with compressed \*.ocf.zip files. When handling a \*.ocf.zip file, this tooling will obey the "unknown file" rule above and leave any other files in the zip untouched.
 
 **You can also break out the various OCF Types and OCF Objects and make those available separately via API or other, suitable mechanism. The key part of OCF is the data structure, not the container itself, though, to ensure complete cap table integrity, a full cap table export is recommended.**
 
--- a/docs/schema/enums/AccrualPeriodType.md
+++ b/docs/schema/enums/AccrualPeriodType.md
@@ -12,4 +12,4 @@
 
 **Source Code:** [schema/enums/AccrualPeriodType](/schema/enums/AccrualPeriodType.schema.json)
 
-Copyright © 2022 Open Cap Table Coalition.
+Copyright © 2023 Open Cap Table Coalition.
--- a/docs/schema/enums/AddressType.md
+++ b/docs/schema/enums/AddressType.md
@@ -12,4 +12,4 @@
 
 **Source Code:** [schema/enums/AddressType](/schema/enums/AddressType.schema.json)
 
-Copyright © 2022 Open Cap Table Coalition.
+Copyright © 2023 Open Cap Table Coalition.
--- a/docs/schema/enums/AllocationType.md
+++ b/docs/schema/enums/AllocationType.md
@@ -19,4 +19,4 @@
 
 **Source Code:** [schema/enums/AllocationType](/schema/enums/AllocationType.schema.json)
 
-Copyright © 2022 Open Cap Table Coalition.
+Copyright © 2023 Open Cap Table Coalition.
--- a/docs/schema/enums/CompensationType.md
+++ b/docs/schema/enums/CompensationType.md
@@ -12,4 +12,4 @@
 
 **Source Code:** [schema/enums/CompensationType](/schema/enums/CompensationType.schema.json)
 
-Copyright © 2022 Open Cap Table Coalition.
+Copyright © 2023 Open Cap Table Coalition.
--- a/docs/schema/enums/CompoundingType.md
+++ b/docs/schema/enums/CompoundingType.md
@@ -12,4 +12,4 @@
 
 **Source Code:** [schema/enums/CompoundingType](/schema/enums/CompoundingType.schema.json)
 
-Copyright © 2022 Open Cap Table Coalition.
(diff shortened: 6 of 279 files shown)


dan-pulley:
- keep the files in the zip
- encourage "lite" exports that dont have files needed for perfect reproduceability
- allow URL-type references to the external files
- include OCF *inside* a larger export

required file retention is problematic for tooling

time to collect thoughts on whether OCF should be narrowly scoped or allow for extra, uncategorized files

pjohnmeyer:
Our team here had a conversation about this topic and here are my notes. I'll state them in terms of being "OCF Compliant" -- which is a term we have [thrown around](https://www.opencaptablecoalition.com/format) but not yet really defined. I'm not saying these are precise wordings we should adopt, but I think we may want to start thinking about rules in this SHOULD / MAY / MUST (+NOT) language. Hopefully these thoughts will help us come to some agreement.

Key definition: I will differentiate between "unreferenced" files -- those that are not listed in the Manifest, and "referenced" files, those that we propose be listed in the manifest.

## 1. Allow extra unreferenced files?
We would be supportive of:
> OCF Compliant software MAY allow unreferenced files as input, or produce unreferenced files as output, in an OCF Container.

We would also be supportive of the following more restrictive wording, if combined with proposal 3 for allowing _specified_ files:
> OCF Compliant software MAY allow unreferenced files as input in an OCF Container. OCF Compliant software MUST NOT produce unreferenced files as output in an OCF Container.

Either of these would be paired with:
> OCF Compliant software MAY restrict size of containers it accepts as input.

Reasoning: Allowing extra files in / out adds the flexibility requested, while the MAY statements prevent creating an onerous requirement on all processors.

## 2. What can software do with unreferenced files?

> If OCF Compliant software allows unreferenced files as input in an OCF Container, it MUST NOT alter the contents of those files when producing a Container derived from the original, _unless_ altered by the end user via the same software.

I think we all agreed to this one on prior calls. The wording of the "unless" part is clunky but it means that if I as the user upload a package from Tool A to Tool B, and then within Tool B -- e.g. some web service -- I somehow update "stuff", then that's allowed.

> If OCF Compliant software does allow unreferenced files as input in an OCF Container, it MAY remove those files before producing a Container derived from the original.

Removal is non-destructive; the user presumably should still have the files and can re-add them to the Container if need be. That's not super user friendly but again, I don't think we want to put an onus on any one processor. That said, the requirement here is also dependent on what we choose for 1. -- if we choose the "MUST NOT produce" language then this requirement, by extension, should be "MUST remove those files" for logical consistency amongst the rules.

We should maybe have some rules about storage (don't store any longer than necessary) and processing (e.g. you're not allowed to transmit those files to another party without express consent, you're not allowed to data mine them, etc.) -- but that's out of scope for this issue+PR, and it's not like we have some enforcement arm or certification body to do any of this stuff so not sure how far down the rabbit hole we should even go.

## 3. Allow referenced files?

All of the above is specifically in combination with a suggestion that we add an array of _referenced_ files to the manifest. However, these files would not _have to_ "physically" reside in the package. They would be specified as an array like such:

```json
[
  {
    "uri": "https://example.com/path/to/file?key=134ab3412523afcc",
    "expires_at": "2023-02-23T01:23:45-06:00"
   },
   {
    "uri": "https://example.com/path/to/another/file",
    "expires_at": "never"
   }
]
```

URIs in referenced files could be true "permalinks" (second entry) but would more likely be some kind of pre-signed link with a limited access window.

> OCF Compliant software MAY include additional "referenced files" as output in an OCF Manifest File.
> OCF Compliant software MUST NOT alter or remove "referenced files" provided as input when producing a new manifest file derived from the input, except as necessary to extend the life of the reference.

This lightweight way of passing along interesting information allows processors to preserve bandwidth and processing time -- only fetching files if absolutely necessary. It also places the InfoSec onus where it already lies -- the user and the originating system -- rather than imparting it on every tool in the pipeline. (With the exception of protecting the links of course.)

One major drawback of this proposal is that an end user may not realize that there are expiring URLs in their final package and they let those links expire. If we extend the concept to allow the referenced file to either be present in the package or not by adding an optional `filepath`, then that resolves that problem, but then it starts to be self-defeating. I haven't thought through how to resolve that issue in a satisfactory way.

jacobyavis:
I'm not in particular favour of supporting unreferenced files.

I do like Patrick's proposal for referenced files, either by URL (lightweight) or by path. It makes sense to me that everything inside the OCF ZIP container is noted in the manifest. 

Certainly if there are documents, images, etc. that are pertinent to the cap table, it makes sense to have them in the container, but they need to be noted.

While this is a great point against allowing local references by path:
> This lightweight way of passing along interesting information allows processors to preserve bandwidth and processing time -- only fetching files if absolutely necessary. It also places the InfoSec onus where it already lies -- the user and the originating system -- rather than imparting it on every tool in the pipeline. (With the exception of protecting the links of course.)

... I think it might be a bit of a hurdle to force implementers to provide these extras by URL.

I'm imagining something that allows reference and _object linking_ as well, (optional `object_id`), so to expand on Patrick's sample from above:

[
  {
    "uri": "https://example.com/path/to/file?key=134ab3412523afcc",
    "expires_at": "2023-02-23T01:23:45-06:00"
   },
   {
    "uri": "https://example.com/path/to/another/file",
    "expires_at": "never"
   },
   {
    "path": "./company_logo",
   },
   {
    "path": "./grant_documents/JohnWilliamsSept2022Grant.pdf",
    "object_id": "security_issuance_1"
   }
]

One caveat is this opens up a manifest file to getting potentially very large depending on what an implementer wants to include (we support splitting everything else into multiple files, but the manifest is a single .json).

One last thought is in addition to being for a referenced file to link to an object this way, we might also want to allow some sort of comments or notes field for these.

dan-pulley:
### Stream of Thought
Here are some initial thoughts and questions:

- Will anyone except the vendor pay attention to untracked files in practice? If so, we need to consider additional_file_type.
- How will URLs work when it comes to opsec? Will they be public? I'm seeing less and less utility in this option.
- I'm only seeing two real use cases for extraneous files. I would love feedback if there are more:
  - Offboarding from a vendor and keeping a reconstructable cap table to onboard again.
  - External tools taking advantage of the included files.

For the first use case:
- The files in the archive are expected to be read only by the initial vendor. They are meaningless to everyone else. Any tooling that changes the OCF won't need to retain those files. So, I wonder if "archive-only" is specifically for vendors to re-onboard?
For the second use case:
- The files in the archive will be utilized but not retained. Essentially, the tooling is mutating the OCF live. In this case, the files would need context about what they are, and they would be "official" files.

Because of this, only files that are explicitly OCF supported have any value to anyone except the vendor. Maybe the solution is simple: we completely ignore extraneous files at all times. Vendors can add another JSON file (without the OCF.json extension) to track and sign their extraneous files. This will be meaningless to anyone else anyway.

The other rule is that files with meaning to OCF get proper schema entries, for example CERTIFICATE_IMAGES.

### Summary Questions

- Is there any value in other tools recognizing the archive files? The other files become "unstable" once the OCF is edited anyway. I feel that an archive can't really be mutated and then re-uploaded losslessly to the initial vendor.
- Is there value in supporting specific file types in the OCF? If that's the case, a rule to always ignore non-OCF files essentially makes an "archive" version equivalent to a normal version except that only the unedited archive version can be re-uploaded to the initial vendor.
- What is the value of an href for files? Why would the vendor want to retain a public link to files forever? This only makes sense for "view-only" OCF tools. For those, it feels overall safer to just include the files instead of having a public link.

### Summary Proposal
In my opinion, other tools will never recognize extraneous files, so we should ignore them. We can slowly add support to specific objects (for example, a share having an option "uri" for the certificate image) with rules around retention, format, and size for those. In that spirit, my current proposal is:

- An "attachments" folder, in which every file MUST be referenced by an OCF object, MUST fit the requirements of that specific file, and MUST be retained by tools that consume or edit an OCF file. We can add "certificate images" to get the ball rolling.
- All other files and folders MUST BE explicitly ignored, not retained by tools, and only useful to original vendors. This is still a valid OCF format (can be uploaded to tools), but the archive is the only file that can reconstruct the cap table on the initial vendor. Vendors are free to add files that help them reconsume an archive, such as a signature of all the attachments. In practice, vendors would likely add a folder like "pulley_attachments" and put all the unsupported files in there, along with signed hashes to verify the archive has remained unedited

JSv4:
Converted to draft as, per our discussion on TWG call today, 3/7/2023, we want to focus our efforts on critical bug fixes, OCX-related changes and validation work

jacobyavis:
Per TWG call on March 21 2023, we will be closing this for now to focus on other things

PR #371, opened 2023-02-06
Added new script generate-release-docs

xudiera:
#### What type of PR is this?

/kind feature

New script to generate the docs with the url on the $id and $ref fields pointing to https://schema.opencaptablecoalition.com/v/M.m.p


package.json
```diff
@@ -6,6 +6,7 @@
     "lint": "prettier . --check --ignore-unknown",
     "fix": "prettier . --check --write --ignore-unknown",
     "generate-docs": "node --loader ts-node/esm --no-warnings --experimental-json-modules ./utils/generate-docs/index.ts",
+    "generate-release-docs": "node --loader ts-node/esm --no-warnings --experimental-json-modules ./utils/schema-utils/GenerateReleaseDocs.ts generate-release-docs -v",
```
pjohnmeyer, line 9: @xudiera I fixed an issue here by removing the `--tag main` from this. I couldn't run the script as desired via `npm run generate-release-docs -- --tag 1.0.1` for example; it would give me refs with `.../v/main,1.0.1/...` in them.

pjohnmeyer: This might imply some needed changes to how we've configured yargs and the `--tag` argument as it shouldn't support multiples, but we can proceed with this.

JSv4 (changes requested):
I'm still investigating the root cause of this, but I'm seeing an issue here. After running `npm run generate-release-docs`, the command to test the example files fails. On investigating, it looks like `URI_LOOKUP_FOR_FILE_TYPE` mapping schema uris to given file types is defined twice - once in validate.mjs and once in Constants.ts in the schema-utils folder. Obviously that should be fixed, but it wasn't causing issues before. With this PR, it causes issues as some of the uris are wrong. Will keep investigating.

utils/validate.mjs
```diff
@@ -7,48 +7,12 @@ import addFormats from "ajv-formats";
 import fs from "fs";
 import { readFile, readdir } from "fs/promises";
 import { resolve } from "path";
+import { URI_LOOKUP_FOR_FILE_TYPE } from "./schema-utils/Constants.ts";
```
JSv4, line 10: This was already defined in "./schema-utils/Constants.ts". Deleted local def.

JSv4:
Thought I was going crazy here... the tests were passing before I deleted the duplicate definition of the OCF file type enum to schema uri mapping in `validate.mjs`. The duplicate definition definitely had the wrong schema root url in the uris (`https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main`)...  According to GitHub, nothing changed in the schemas after running the generate-release-docs command, yet all of a sudden the tests broke when I __removed__ the wrong uris... 

Turns out we approved a PR that changed the $ids to the wrong root: 

image

The issue here we need to fix is two-fold. **First**, we accidentally overwrote the correct $ids. **Second**, after the generate-release-docs command runs, we do not currently update the mapping between the OCF File Type enums and their corresponding uris that's currently defined as an object in `./utils/schema-utils/Constants.ts`:

image

JSv4:
Per the call today, we actually do want to use the GitHub raw root url. My bad. We DO need to update the file type to uri mapping dynamically as part of generate-release-docs though.

JSv4:
@xudiera, so here's the behavior I'm seeing. If you start with a fresh copy of this branch (just started on a totally fresh machine), run 

`npm run validate-ocf-file-schemas` - should pass
`npm run validate-example-ocf-files` - should pass

Now, run your script with `npm run generate-release-docs`. Then run `npm run validate-example-ocf-files` and you get an error:

image

Like I said, maybe we're ok with this behavior? Personally, I think the tests should still after generating the release versions of the schemas. I believe the reason the test is not working is the mapping between the ocf file type enumeration and the ocf schema $id is not being updated along with the schemas themselves.

JSv4:
Any updates here, @xudiera?

xudiera:
> Any updates here, @xudiera?

Sorry, I was on vacation. I just pushed the fix to the issue you previously reported.

Now running `npm run validate-ocf-file-schemas` and `npm run validate-example-ocf-files` after `npm run generate-release-docs` doesn't trigger an error anymore. I moved the URIs for the file types lookup to `utils/schema-utils/UriLookupForFileType.json`.

JSv4 (approved):

PR #378, opened 2023-02-22
Suggested refractor for easier code generation

regebro:
#### What type of PR is this?

/kind feature

When generating Python code from the schema, I encountered a problem. Several types reference primitives to implement those schemas. In Python, using the primitives as a baseclass has several benefits. For example you can check if something is a ConversionMechanism by doing `isinstance(object, ConversionMechanism)` with no need to list all possible types of conversion mechanisms.

However, the primitive types reference instantiatable types, and this causes circular imports in Python. Basically when a type using a ConversionMechanism gets loaded this leads to loading of the primitive, that leads to loading the other primitives, which tries to load the instantiable types, one of is currently being loaded, and therefore it fails.

By referencing only primitive types from primitives we avoid these circular imports.

There are various other ways of avoiding this (at least in Python, and this may very well be a Python only problem), but they all result in "ugly" code, one way or another, so this is not a deal breaker. But this is my preferred solution.

Avoids circular imports in Python in a nice way.

### DIFF
--- a/schema/primitives/types/conversion_rights/ConversionRight.schema.json
+++ b/schema/primitives/types/conversion_rights/ConversionRight.schema.json
@@ -11,26 +11,7 @@
     },
     "conversion_mechanism": {
       "description": "What conversion mechanism applies to calculate the number of resulting securities?",
-      "oneOf": [
-        {
-          "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/types/conversion_mechanisms/SAFEConversionMechanism.schema.json"
-        },
-        {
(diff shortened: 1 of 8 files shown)


JSv4:
Thanks @regebro, we'll take a look at the proposal.

JSv4:
Hey @regebro, can you join the next Technical Working Group call at Noon Eastern on the 28th? It would be great to hear how you're thinking about this and maybe walk through some of the downstream effects our current structure has on your Python tooling.

regebro:
Sorry to say, in the move over to Fidelity systems, I have lost the meeting link.

regebro:
@JSv4 Sorry to say, in the move over to different systems, I have lost the meeting link.

pjohnmeyer:
**TWG** Closing in favor of future resolution of #405

PR #380, opened 2023-02-28
Added proposal to give stock plans a clear default rule on what happe…

### LINKED ISSUE #370: [Enhancement]: Explore Way to Capture When and Whether Option Shares Go Back Into Pool

Shares of stock reserved for options don't always go back into the Stock Plan. This varies on a plan-by-plan basis. Let's explore if there are consistent types of events where this distinct is drawn and see if we can add a data structure to the Stock Plan obj to capture this at the plan label. 

### Why is this Needed?

The rules here - do the option shares go back into the pool or not - are not consistent from plan to plan. One platform may assume certain things that others do not assume (or make explicit to the user).

JSv4:
#### What type of PR is this?

/kind feature

As discussed in the TWG, this is one of two proposed changes to make it explicit what happens to securities when they're cancelled. Added a `cancellation_behavior` field to `Stock Plan` to hold an enum specifying what should happen to stock after the plan securities are cancelled.

Fixes #370


schema/enums/StockPlanCancellationBehaviorType.schema.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/enums/StockPlanCancellationBehaviorType.schema.json",
+  "title": "Enum - Stock Plan Cancellation Behavior Type",
+  "description": "For a given stock plan, what is the default rule for what happens to the shares reserved for a Plan Security after it's cancelled.",
+  "type": "string",
+  "enum": ["RETIRE", "RETURN_TO_POOL", "CAPITAL_STOCK"],
```
JSv4, line 7: Add an option for per_security

JSv4: We should also update the validation meta issue with this rule - what implications per_security has.

JSv4:
FYI, I am seeing an issue with the documentation generator... I thought this was fixed as part of an earlier PR (maybe it wasn't merged yet, I have to go look) ... but looks like we need to update how we're handling certain imports due to our upgraded node version: 

image

JSv4:
> FYI, I am seeing an issue with the documentation generator... I thought this was fixed as part of an earlier PR (maybe it wasn't merged yet, I have to go look) ... but looks like we need to update how we're handling certain imports due to our upgraded node version:
> 
> image

Yeah, looks like my local documentation isn't updating. I need to trace this and figure out if it's something in my environment or if we have a PR that needs merging or authoring. This is definitely related to node version though as I recall seeing this earlier.

schema/objects/StockPlan.schema.json
```diff
@@ -29,6 +29,10 @@
       "description": "The initial number of shares reserved in the pool for this stock plan by the Board or equivalent body.",
       "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/types/Numeric.schema.json"
     },
+    "cancellation_behavior": {
```
jacobyavis, line 32: We should consider naming this `default_cancellation_behavior` if we allow it to be overridden on a transaction level (see PR #381)

JSv4:
Thinking about this a little bit more... the corollary change to this PR is the creation of specific events showing what happened to shares reserved for a security under a plan - See PR #381.

Thinking about this, though, shouldn't we then create transaction for each of the enumerated defaults? 

For reference current enum proposes:

1. `RETURN_TO_POOL` - Shares return to the pool.
2. `HOLD_AS_CAPITAL_STOCK` - Shares are held by company as capital stock.
3. `RETIRE` - Shares are retired entirely.
4. `DEFINED_PER_PLAN_SECURITY` - No plan-wide default and this is handled on a Plan-Security-by-Plan-Security basis.

We have a return to pool event defined. `PlanSecurityCancellation` (or whatever we end up calling it) doesn't cleanly map to the `RETIRE` option, though, because that describes cancelling the security issued under the plan, NOT cancelling the reserved shares. Likewise, we don't have a clean event for making something capital stock. Repurchase... maybe, but, I don't think it's clean enough. I think we need two more event types - a TX_[ ]_share_cancellation_event and a TX_[]_make_shares_capital_stock event. Thoughts?

pjohnmeyer:
@daviidy note this for potential OCX calculation impact

PR #381, opened 2023-02-28
[Enhancement] - Add draft transaction to return plan securities' shares to pool

### LINKED ISSUE #370: [Enhancement]: Explore Way to Capture When and Whether Option Shares Go Back Into Pool

JSv4:
#### What type of PR is this?

/kind feature

This is the first of three, related transactions proposed to clearly delineate where plan securities are a) returned to a pool, b) retired or c) held as capital stock. In the case of (b) and (c), it probably makes sense to propose similar transactions for stock. 

Submitted per request from TWG discussion for a proposal. 

Fixes #370


JSv4:
Per TWG call let's make the transaction a Stock Plan transaction and have it reference the event causing the return to pool. Also take a look at validation meta issue and add any implicit rules from this schema into that issue.

JSv4:
As discussed, I made this a transaction that affects the plan. `PlanSecurityReturnToPool` became `StockPlanReturnToPool`. When I added the `StockPlan` transaction primitive to the list of primitives we used to compose this, I noticed it actually didn't change the fields... we already had all of the fields required for a `StockPlan` transaction. I thought it made sense to keep the `PlanSecurity` transaction primitive too because even with refactor, we still need the data fields that would make it a `PlanSecurity` transaction in addition to the fields required for a `StockPlan` transaction... I don't think we have many (any?) objects that are composed of primitives from transactions meant for different object types, but I don't think it's 'wrong'. As this is written, it's is unusual, however. We need the fields it has, though, so the question is really whether we want to remove the primitive for the PlanSecurity transaction or keep it. The fields wouldn't change.

Ready for review in the meantime, @jacobyavis or @pjohnmeyer

pjohnmeyer:
@JSv4 I think this looks good, and I _think_ I like that it references both transaction primitives? Agree we should discuss at TWG.

jacobyavis (approved):
Thanks John, looks good

pjohnmeyer:
@daviidy Note this for OCX calculation impact

PR #392, opened 2023-03-11
Validate obj against obj_type schema

### LINKED ISSUE #388: [Ehancement] - Improve Validator to Look for OCF Type Enum and Only Validate Against Applicable Schema
As @jacobyavis and @naveedn point out, getting more helpful outputs from the json schema validators would be great. Transaction validations in particular are really cryptic because the schema for the transaction file is basically just a `oneOf` list with tons of different schemas that could match. Jsonschema doesn't know that each schema has a type tag and that the type tag is unique per schema. It should check to see if there's an object type tag first, and, if it finds one, only validate against the applicable schema for that object tag. This is something we could build pretty easily - basically, for each validation of an OCF object, first get the `type` tag, then use a lookup table to find the appropriate OCF schema, and then ONLY validate against the applicable schema. We need to build similar tooling to crawl the schemas and build a mapping table from enums to $refs for the changes @xudiera needs to make to #371 too.

_Originally posted by @JSv4 in https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/issues/359#issuecomment-1458469791_

naveedn:
/kind feature

In the current state, the validator behaves like this:
- It ingests every file in a given directory, and compares it to the corresponding file schema
- each file object has an `items` array (with the exception of the manifest file), and in the case of the transactions file, each `item` can be `"oneOf"` many different kinds of schemas.
- the validator will attempt to validate the item being processed against every schema in it's `"oneOf"` list and record the errors for all the schemas.

**Result:** all of the validation errors are shown, instead of the one that is most pertinent to the user. 

#### This PR changes the validator's behavior in the following way:
- Adds a few helper functions that map each schema's `object_type` => `const` value to the schema's `$id`. This is referred to as the `objectSchemaMap`
- when the validator loops through every item in the `items` array:
       - grabs that item's corresponding schema `$id` with the `objectSchemaMap`
       - Grabs the schema associated with the $id from the [ajv instance cache](https://ajv.js.org/guide/managing-schemas.html#using-ajv-instance-cache)
       - Validates the item against the matched schema 

Fixes #388

#### TODO:
- [ ] fix the if / else logic so that file level validation occurs on all files as well as the object specific validation 
- [ ] refactor the code

### DIFF
--- a/utils/validate.mjs
+++ b/utils/validate.mjs
@@ -53,20 +53,16 @@ export const URI_LOOKUP_FOR_FILE_TYPE = {
 async function buildObjectSchemaMap(verbose = false) {
   const schemaMap = {};
 
-  if (verbose) console.log("\n-->\tLoad Schema Files...\n");
   const schemaPaths = await getSchemaObjectsFilepaths();
 
-  if (verbose) console.log("\n-->\tParse Schema Objs...");
   const schema_buffers = await Promise.all(
     schemaPaths.map((path) => readFile(path))
   );
 
-  if (verbose) console.log("\n-->\tParsing Schema JSONs");
   const schemas = schema_buffers.map((schema_buffer) => {
     return JSON.parse(schema_buffer.toString());
   });
 
-  if (verbose) console.log("\n-->\tBuilding Schema Map");
   schemas.forEach((schema) => {
     schemaMap[schema.properties.object_type.const] = schema.$id;
   });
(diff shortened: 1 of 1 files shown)


JSv4 (approved):
This is awesome @naveedn. Love the new validation error message.

naveedn:
fixed the merge conflicts, should be good to go!

PR #396, opened 2023-03-30
Add RSA Support

### LINKED ISSUE #379: [Question]: How to model an RSA / RSA repurchase?
### Question:

It's unclear to me how a Restricted Stock Award issuance and/or repurchase is correctly modeled in OCF. Is there a way, or are we missing something from the schema?

1. The only reference to an RSA I can find in the schema documentation is in the documentation for the [StockParent](https://open-cap-table-coalition.github.io/Open-Cap-Format-OCF/schema_markdown/schema/types/StockParent/) type. I cannot find any references to this type in the schema.

2. There is no mention of "types" of stock awards in [StockIssuance](https://open-cap-table-coalition.github.io/Open-Cap-Format-OCF/schema_markdown/schema/objects/transactions/issuance/StockIssuance/), so it is unclear how somebody could distinguish between a regular stock issuance and an RSA issuance.

3. I don't know if it would be appropriate to use a [PlanSecurityIssuance](https://open-cap-table-coalition.github.io/Open-Cap-Format-OCF/schema_markdown/schema/objects/transactions/issuance/PlanSecurityIssuance/) to describe an RSA issuance, but if it were, currently `compensation_type` only lists `OPTION` and `RSU`. Even if an RSA could be correctly modeled as a PlanSecurityIssuance, there is no corresponding PlanSecurityRepurchase.

JSv4:
#### What type of PR is this?

/kind bug

We have an ongoing dialog (and more changes to come) to better support the grab bag of compensation-related equity that's out there and all of the corner-case wrinkles - plan-less options, stock issued from plans, SARs, etc. Specifically for RSAs, it was our conclusion that the existing Stock objects (create on the Stock Issuance) have everything required for an RSA except for the possibility of being linked to a plan. This change adds an optional stock plan id field to the stock issuance so you can create an RSA. Note currently I'm not proposing any kind of flag or formal notation this is an RSA. I don't believe that's strictly required. 

This is technically a bug fix as the Plan Securities objects don't currently include RSAs and there's no way to issue vesting stock awards linked to a plan.

Fixes #379  This isn't strictly supporting RSA repurchases, but it enables a clean representation of RSAs which can then be repurchased like any other type of Stock.


jacobyavis (approved):

JSv4:
From the TWG Call:

Let's try to put something to make it explicit when this is RSA. Considering an Enum at the moment, which gives us an "upgrade" pathway for future values and also could help identify things like founders stock more explicitly.

schema/enums/StockIssuanceType.schema.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/enums/StockIssuanceType.schema.json",
+  "title": "Enum - Stock Issuance Type",
+  "description": "Enumeration of issuance types where we want to draw attention to some unique aspect of a stock issuance (e.g. is it an RSA)",
+  "type": "string",
+  "enum": ["RSA"],
```
JSv4, line 7: Also add founders' stock.

PR #397, opened 2023-03-31
Revise PlanSecurity to Permit Plan-less Options and Incorporate Additional Type Enums

JSv4:
#### What type of PR is this?

/kind bug

As discussed on the TWG, we need to support plan-less options. We also needed to support stock issuances under a pool with vesting for RSAs. As discussed, we also want to support SARS. Other synthetic "equity" may be of interest now and in the future too ("Phantom Stock" for one). A challenge that is immediately obvious is shoe-horning all of these things into four equity types is getting increasingly tenuous and the edges are bleeding. Just one simple and obvious example - most people see RSAs as equity compensation BUT they are structurally identical to stock with vesting and an optional stock plan link. Discussed this with some folks on our KM team and with @pjohnmeyer and some kind of longer-term refactor to break out every discrete type of security into its own separate object is probably advisable. Something like this:

Long-Term Refactor Proposal for v2.0.0
image

This is a huge change, however. There is a short-term fix that can be structure to avoid a breaking change:

Immediate Refactor Proposal to Fix Equity Compensation without Breaking Changes
image

The proposed PR achieves option 2 (in connection with PR #396).

Note the full changes required:
- Renamed PlanSecurity to EquityCompensation (otherwise it's kind of weird with plan-less PlanSecurities of type "option"
- Included all of @jacobyavis' enums minus RSAs (see PR #396)
- Added a grant price for SARs. Added validations to make it required for SARs
- Used a oneOf in the object_type field to support old `TX_PLAN_SECURITY_*` objects and new `TX_EQUITY_COMPENSATION_*` objects. 
- Updated document generator to handle the new design pattern for `oneOf` in the object_type field, which we've never used before. 

TODO - I need to make 100% sure the doc generator tests still pass. As pointed out by @xudiera, they don't trigger the actions to fail if the tests fail. I think they should fail given the MD output is now slightly different for the new EquityCompensation schemas.

Fixes #


schema/enums/ObjectType.schema.json
```diff
@@ -22,6 +22,13 @@
     "TX_CONVERTIBLE_ISSUANCE",
     "TX_CONVERTIBLE_RETRACTION",
     "TX_CONVERTIBLE_TRANSFER",
+    "TX_EQUITY_COMPENSATION_ACCEPTANCE",
+    "TX_EQUITY_COMPENSATION_CANCELLATION",
+    "TX_EQUITY_COMPENSATION_EXERCISE",
+    "TX_EQUITY_COMPENSATION_SSUANCE",
```
jacobyavis, line 28: typo here on ISSUANCE

schema/objects/transactions/issuance/EquityCompensationIssuance.schema.json
```diff
@@ -20,10 +20,18 @@
   ],
   "properties": {
     "object_type": {
-      "const": "TX_PLAN_SECURITY_ISSUANCE"
+      "oneOf": [
+        {
+          "const": "TX_PLAN_SECURITY_ISSUANCE"
+        },
+        {
+          "const": "TX_EQUITY_COMPENSATION_ISSUANCE"
```
jacobyavis, line 28: no typo here though, so this would cause the schema validator to fail

schema/enums/ObjectType.schema.json
JSv4: Thx. Fixed

schema/enums/ObjectType.schema.json
```diff
@@ -22,6 +22,13 @@
     "TX_CONVERTIBLE_ISSUANCE",
     "TX_CONVERTIBLE_RETRACTION",
     "TX_CONVERTIBLE_TRANSFER",
+    "TX_EQUITY_COMPENSATION_ACCEPTANCE",
+    "TX_EQUITY_COMPENSATION_CANCELLATION",
+    "TX_EQUITY_COMPENSATION_EXERCISE",
+    "TX_EQUITY_COMPENSATION_ISSUANCE",
```
JSv4, line 28: Fixed here...

jacobyavis (approved):

pjohnmeyer:
@JSv4 I don't see a way to determine what stock class non-plan awards convert to in this schema.

JSv4:
> @JSv4 I don't see a way to determine what stock class non-plan awards convert to in this schema.

Thanks @pjohnmeyer. I knocked out a couple of long-pending issues this week and merged some PRs. I like your proposed fix for the issue with the `object_type` property.

Good catch on the stock class for non-plan awards. I will incorporate all of your suggestions, hopefully today. Converted this to a draft in the meantime.

JSv4:
> @JSv4 I did some experimenting locally and I think there's a smaller change-set here that uses an `enum` for `object_type` instead of `oneOf[*].const`. It does unfortunately ripple into the document generator, the same way your changes did, but the enum is a little easier to read. The steps look like:
> 
>     1. Change PlanSecurity* txns to have `object_type: { enum: ["TX_EQUITY_COMP...", "TX_PLAN_SEC..."] } }`
> 
>     2. Copy each PlanSecurity* txn to the new EquityCompensation* schema files and re-id
> 
>     3. Change each PlanSecurity* txn to _only_ be: `allOf: [ { $ref: "https://.../EquityCompensation..." } ]`
> 
> 
> While we don't have this written down, when previously discussing what would be a breaking change, we discussed the balance of defining what the "interface" is, and being able to work and iterate freely "behind" that interface. At that time, I proposed that deleting or renaming an **_object_** should be considered breaking -- but that we should be able to rename / reorganize _enums, types, primitives_, etc., as being the "private" types of the schema.
> 
> I think it is a reasonable expectation of developers that they should be able to reference an $id for validators such as `https://schema.opencaptablecoalition.com/v/1.0.1/objects/Issuer.schema.json` -- and then when they update to 1.0.2 just change the version string. Having to know that an object was renamed is a developer experience problem, especially when writing otherwise version-agnostic bits of code. It is true that your changes here are not breaking when validating an entire OCF package, but it somebody were, for example, implementing an API against specific objects and validating directly against PlanSecurityIssuance.schema.json, it would be breaking.
> 
> I can continue to work on my `enum` version of this between now and next working group and hopefully have it at least partially done before then, or we can reinstate the PlanSecurity* types on this PR as deprecated aliases to the new objects, but I think this is a reasonable line to draw. What do you think?

Agree with your reasoning here. Changing the $id is a breaking change for certain consumers of the standard, so your proposed approach would avoid that. I hope we can avoid doing this again, however, as it's a whole mess of duplication that I wish we didn't have.

pjohnmeyer:
> Changing the $id is a breaking change for certain consumers of the standard

Can we agree this only applies to `files` and `objects` though? If so I can write up a "version policy" doc.

pjohnmeyer:
@JSv4 one follow-up on my own comment:

> I think there's a smaller change-set here that uses an enum for object_type instead of oneOf[*].const

In #408 I actually used an `enum` and a `oneOf` until we have an opportunity to discuss the approach I took. The nice thing about `oneOf` instead of `enum` is that each option can have its own `description`.

schema/files/TransactionsFile.schema.json
```diff
@@ -55,7 +55,7 @@
             "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/objects/transactions/issuance/ConvertibleIssuance.schema.json"
           },
           {
-            "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/objects/transactions/issuance/PlanSecurityIssuance.schema.json"
```
JSv4, line 58: @pjohnmeyer, what do you think, should we keep both $ids in the TransactionsFile items permitted $refs? Seems like yes, even if, for validation purposes, I don't think it should be strictly necessary. Someone may be referencing or parsing this list and be surprised by the omission of the PlanSecurity... $ids.

utils/generate-docs/Schema/SchemaNode/Property/InlineProperty/OneOfArray.ts
```diff
@@ -1,16 +1,17 @@
 import SchemaNode from "../../SchemaNode.js";
 import PropertyFactory, { PropertyJson } from "../Factory.js";
-import InlineProperty from "./InlineProperty.js";
+import InlineProperty, { InlinePropertyJson } from "./InlineProperty.js";
 
 interface Schema {
   findSchemaNodeById: (id: string) => SchemaNode;
 }
 
-export interface OneOfArrayJson {
```
JSv4, line 9: I had made this change when I thought we'd use a oneOf array of {"const":[obj_type]} for the `object_type` fields. Let's us dynamically change the typing in the oneOf items list. I figured I'd keep this change here as it may be useful and doesn't break anything.

JSv4:
> @JSv4 one follow-up on my own comment:
> 
> > I think there's a smaller change-set here that uses an enum for object_type instead of oneOf[*].const
> 
> In #408 I actually used an `enum` and a `oneOf` until we have an opportunity to discuss the approach I took. The nice thing about `oneOf` instead of `enum` is that each option can have its own `description`.

Sounds good. You can also add $comments to them which we might consider using for deprecation warnings. 

> > Changing the $id is a breaking change for certain consumers of the standard
> 
> Can we agree this only applies to `files` and `objects` though? If so I can write up a "version policy" doc.

Yes, 100%. Thanks for offering to write something up.

JSv4:
> @JSv4 one follow-up on my own comment:
> 
> > I think there's a smaller change-set here that uses an enum for object_type instead of oneOf[*].const
> 
> In #408 I actually used an `enum` and a `oneOf` until we have an opportunity to discuss the approach I took. The nice thing about `oneOf` instead of `enum` is that each option can have its own `description`.

Cool. Yeah, that was the approach I took in the last iteration. Could be convinced to go back. Another pro was you can have $comment props on each of the `oneOf` constants, and we could consider using the comment field for warnings / messages that are more significant than what would normally go in the description - e.g. documentation versus a deprecation warning - the warning maybe would go in the $comment field.

schema/files/TransactionsFile.schema.json
pjohnmeyer: If it validates either way I'm not sure I care.

PR #399, opened 2023-04-03
Removed test-script.js

### LINKED ISSUE #368: [Bug]: Fix Test Not Triggering Failure in GitHub Action
### Current Behavior:

Here is an example of the Action not failing when a unit test doesn't pass:
https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/actions/runs/3807911702/jobs/6478044986#step:5:43

I think this can be solve by  set the Fail-fast behavior on [jest.yml](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/main/.github/workflows/jest.yml)

```
- name: Run Jest Tests
        run: npm run test
        shell: bash
```

See:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

https://github.com/github/docs/blob/2ec9640ca001ef7601c61891cf4538ab220e7961/content%2Factions%2Fusing-workflows%2Fworkflow-syntax-for-github-actions.md#example-running-a-script-using-bash

_Originally posted by @xudiera in https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/issues/355#issuecomment-1381042854_

### Expected Behavior:

If a test fails, GitHub action should fail too.

### Steps to Reproduce:

See above

### Environment:

GitHub Action VM

xudiera:
#### What type of PR is this?

/kind bug

The GitHub action is not failing when a unit test is failing. This PR is solving this be removing the `test-script.js` file. This will prevent running tests on Windows. Currently the project has so many issues in Windows that my recommendation to a developer with a Windows OS will be to do the development using [WSL](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux) instead.

Fixes #368


JSv4:
Let's add some documentation as well to help people get Visual Studio Code running with WSL and then this is great. Thanks!

JSv4 (approved):
LGTM

PR #400, opened 2023-04-04
Update to the contact info type inside the Stakeholder object

### LINKED ISSUE #356: [Enhancement]: Add optional email(s) and phone number(s) variables to stakeholder object

Add optional Email(s) field to the stakeholder object.

### Why is this Needed?

Right now, the only way to add an email and/or phone number to a stakeholder object is through the primary contact type in the stakeholder object and the primary contact type requires name, phone and email.

Primary contact makes sense as is for Institutional stakeholders but less sense of individual stakeholders. Having optional emails & phone numbers fields directly for individual stakeholders makes more logically sense.  

An alternative is to use the primary contact type but not make both phone and email required. Or if required only means you need the variable and can input an empty array, then it is less of an issue.

arthur-clara:
#### What type of PR is this?
/kind feature

Creates a new contact info type for individual stakeholders that does not require name

Fixes #356


jacobyavis (approved):

JSv4 (approved):
LGTM. Thanks, @arthur-clara

PR #410, opened 2023-04-20
Allow for Warrant Issuances with No Quantity or Estimated Quantities and Support Valuation-Based Warrant Exercise Calculations

### LINKED ISSUE #362: [Enhancement]: Enhance Warrant Conversion Spec

A couple, inter-related limitations to be addressed here:

1. Our warrants *require* an exercise price at issuance, which is not always known. Make this optional
2. Our conversion mechanisms for warrants are pretty limited. Reflect some of the mechanisms we have for convertibles for warrants as well. 

### Why is this Needed?

This would allow for better coverage of real-world Warrant terms in our JSON schema. 

Not at the moment.

### LINKED ISSUE #372: [Enhancement]: Proposal to Tweak CustomConversionMechanism to Provide for Min/Max

Legal working group suggested a common use case for warrant exercise is to have some kind of formula-based exercise (which would go under out CustomConversionMechanism) that has clear min and max values. Currently we have *No* structure at all on the [schema/types/conversion_mechanisms/CustomConversionMechanism](https://open-cap-table-coalition.github.io/Open-Cap-Format-OCF/schema/types/conversion_mechanisms/CustomConversionMechanism.schema.json). At a minimum, may want to add min/max OR create a separate conversion mechanism. One possible danger here, though, is platforms probably don't have this data, so that creates mapping ambiguity or outright incompatibility .

### Why is this Needed?

Proposed that this information be available in OCF as lawyers need this for tie-outs and pro formas. 

To share some relevant language for this.

JSv4:
#### What type of PR is this?

/kind bug

As discussed in #372, there is a desire to have some concrete value for warrant conversions in OCX. For modeling purposes a max or an estimated value is OK, so long as it's clear when the quantity is not fixed in the instrument itself. We will support Warrants missing an estimate or known value in OCX, however, as not all platforms collect estimated values and not all warrants have fixed quantities. They'll go in a table where you can provide your own estimates within OCX for modeling purposes. 

As discussed in #362 (and related to the points raised above), Warrants often don't have a fixed exercise price. Related to the above, the quantity of resulting shares may not be fixed and may be dependent on the company's valuation and/or capitalization like SAFEs.  Our current warrant issuance requires an exercise price, however, and doesn't support valuation-based exercise calculations like SAFEs.  

This PR makes a few changes to address these two issues:

- `exercise_price` is no longer required on warrant issuances
- `quantity` is no longer required on warrant issuances
- `quantity_source` allows you to specify where the quantity came from using `QuantitySourceType` enum. For now, it's implicitly `UNSPECIFIED` where `quantity` is provided but `quantity_source` is not. I'd like to require `quantity_source` where `quantity` is provided in the future (v2.0.0). 
- There is now a `VALUATION_BASED_CONVERSION_MECHANISM` which has a number of the fields from the `SAFE_CONVERSION_MECHANISM` related to the value of the company / size of the round. This `VALUATION_BASED_CONVERSION_MECHANISM` is now permitted for warrants. Could make sense to make it available on other instruments too.  

Fixes #372 and Fixes #362


pjohnmeyer (approved):

pjohnmeyer:
Looks good to me but let's go over it in TWG next week.

docs/schema_markdown/schema/types/conversion_mechanisms/ValuationBasedConversionMechanism.md
JSv4, line 1: Discounts are not something that anyone has really seen. We can remove that field. We also want to tweak the valuation cap field to account for situations where the exercise may be based on one or the other or both, depending on the circumstances.

JSv4:
Let's split out the quantity type from the work on the valuation-based conversion mechanism.

JSv4:
Closed as this was split into #426 and #427

Issue #413, opened 2023-04-24
[Enhancement]: Compatibility with jsonschema2pojo


jacobyavis:
### Description of Enhancement :

To be able to create java classes for the OCF schema using popular library [jsonschema2pojo](https://www.jsonschema2pojo.org/)

### Why is this Needed?

Without the `javaType` property on the schemas, the library ends up creating duplicate classes. It does not know that two different references to Stakeholder are supposed to be the same object. 

Here's an example screenshot of how, even with proper $refs, jsonschema2pojo generates a local `Issuer.java` inside `schema/files` instead using schema/objects/Issuer.java

I also tried flattening the schema hierarchy before generating (not ideal), but I still got some duplicates files ending in __1, and also some duplicate enum classes created as inner classes...

### Anything else we need to know?

After reading some forums it seemed the easiest solution is to add `javaType` to the schemas. Tested and it works great so we just need to enforce it using a husky hook

[Reference in the docs](https://www.jsonschema2pojo.org/#:~:text=Extensions-,javaType,-jsonschema2pojo%20supports%20an)

pjohnmeyer:
> Here's an example screenshot

@jacobyavis I don't see the screenshot?

regebro:
Wouldn't the reasonable way to use this schema from Java be to use a common library? It seems strange to me that everyone using this schema would generate their own classes, but maybe that's because I'm from the Python world.

jacobyavis:
Code generation is pretty common in Java projects. At least in my opinion, it's much quicker to run a code-gen lib like `jsonschema2pojo` than to work on maintaining a library in parity with the schema.

PR #414, opened 2023-04-24
add new script to add 'javaType' property to all schemas, for compatibility with jsonschema2pojo

jacobyavis:
#### What type of PR is this?
/kind feature

This adds jsonschema2pojo compatibility as detailed in issue #413

### DIFF
--- a/utils/schema-utils/AddJavaTypeToSchemas.ts
+++ b/utils/schema-utils/AddJavaTypeToSchemas.ts
@@ -1,0 +1,115 @@
+import * as fs from "fs";
+import { readFile } from "fs/promises";
+
+import yargs, { Arguments } from "yargs";
+import { hideBin } from "yargs/helpers";
+
+import { getSchemaFilepaths } from "./Loaders.js";
+import { basenameRelativePathToSchemaDir } from "./PathTools.js";
+
+/**
+ * Given a schema with a known local schema_path, add a 'javaType' property for use by jsonschema2pojo
+ * @param schema_path -> Local path where schema_obj was loaded from. WILL BE OVERWRITTEN with serialized data from updated schema_obj.
+ * @param schema_inst -> OCF schema JSON to update.
+ * @param verbose -> Boolean: Display detailed logs.
+ */
+export function addJavaTypePropertyToSchema(
+  schema_path: string,
+  schema_inst: Record<string, any>,
+  verbose: boolean = false
+) {
+  let java_type_contents = `com.ocf.schema.${basenameRelativePathToSchemaDir(
+    schema_path
+  )}`;
+  java_type_contents = java_type_contents.replaceAll("/", ".");
+  schema_inst["javaType"] = java_type_contents;
+  let schema_data = JSON.stringify(schema_inst, null, 2);
+  fs.writeFileSync(schema_path, schema_data);
+  if (verbose)
+    console.log(
+      `\t--> Inserted valid 'javaType' field in schema ${schema_inst["$id"]}`
+    );
+}
+
+/**
+ * Load all schema files in the /schemas folder. Writes the 'javaType' field to all schemas
+ * e.g. 'com.ocf.schema.objects.Issuer'
+ * This enables OCF for use with jsonschema2pojo
+ *
+ * @param verbose - boolean. If true, display verbose test and result messages to console.
+ * @returns Promise<void>
+ */
+export async function writeJavaTypeToSchemas(
+  verbose: boolean = false
+): Promise<void> {
+  if (verbose) {
+    console.log(`ADD JAVA TYPES ----------------------`);
+  }
+  try {
+    // Load Schema JSONs from file contents
+    const schema_paths = await getSchemaFilepaths(verbose);
+    const schema_buffers = await Promise.all(
+      schema_paths.map((path) => readFile(path))
+    );
+    const schemas = schema_buffers.map((schema_buffer, index) => {
+      if (verbose)
+        console.log(
+          `\t(${index}) Parse JSON Schema at path ${schema_paths[index]}`
+        );
+      return JSON.parse(schema_buffer.toString()) as Record<string, any>;
+    });
+
+    if (verbose)
+      console.log(
+        `Enforce 'javaType' properties on ${schemas.length} total schemas:`
+      );
+
+    // Loop over the schemas to write 'javaType' fields
+    for (let i = 0; i < schemas.length; i++) {
+      let schema_inst = schemas[i];
+
+      if (verbose)
+        console.log(`•\tWriting 'javaType' field for ${schema_inst["$id"]}`);
+
+      addJavaTypePropertyToSchema(schema_paths[i], schema_inst, verbose);
+    }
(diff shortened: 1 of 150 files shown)


pjohnmeyer:
I'm not a huge fan of this; based on the issue it seems like we're fixing a bug in `jsonschema2pojo` by changing our schema files.

It does make me think that maybe we should move our "raw" schema definitions into another set of files entirely, then generate the real schema files with all of the "always" stuff like $schema, $id, $ref, $comment, additionalProperties: false ...

JSv4:
You're thinking some kind of scaffolding tool, basically, where we auto-generate new schemas rather than incorporate the properties via $ref?

jacobyavis:
I think this is a good idea. To introduce almost like a 'build' step that transforms the raw schema src into the finished schema.

So given a src file like so: 
```{
  "$id": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/types/Email.schema.json",
  "title": "Type - Email",
  "description": "Type representation of an email address",
  "type": "object",
  "properties": {
    "email_type": {
      "description": "Type of e-mail address (e.g. personal or business)",
      "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/enums/EmailType.schema.json"
    },
    "email_address": {
      "description": "A valid e-mail address",
      "type": "string",
      "format": "email"
    }
  },
  "required": ["email_type", "email_address"]
}
```

The build step would generate a finished schema like so:
``` {
  "$id": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/types/Email.schema.json",
  "title": "Type - Email",
  "description": "Type representation of an email address",
  "type": "object",
  "properties": {
    "email_type": {
      "description": "Type of e-mail address (e.g. personal or business)",
      "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/enums/EmailType.schema.json"
    },
    "email_address": {
      "description": "A valid e-mail address",
      "type": "string",
      "format": "email"
    }
  },
  "required": ["email_type", "email_address"],
  "additionalProperties": false,
  "$comment": "Copyright © 2023 Open Cap Table Coalition (https://opencaptablecoalition.com) / Original File: https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/tree/main/schema/types/Email.schema.json",
  "javaType": "com.opencaptablecoalition.ocf.schema.types.Email"
}
```

One benefit this brings us is being able to ignore the generated sources, so PRs like this one (and many others, are not difficult to review, 146 schema files!)

I'd be interested in tackling this as a separate issue, but I'm still hoping to get `javaType` in as is for the next release, so that the attribute will appear on the hosted schemas at `raw.githubusercontent.com`

jacobyavis:
To expand on the idea a bit further, it would be nice if the generated schemas which use Composition had the properties expanded out further.

Currently jsonschema2pojo treats anything listed with empty braces as an unknown type, generic `Object`, so I already am planning to tackle this in some way. Perhaps this makes the most sense to do all together

utils/schema-utils/AddJavaTypeToSchemas.ts
```diff
@@ -0,0 +1,115 @@
...
+ * @param schema_path -> Local path where schema_obj was loaded from. WILL BE OVERWRITTEN with serialized data from updated schema_obj.
+ * @param schema_inst -> OCF schema JSON to update.
+ * @param verbose -> Boolean: Display detailed logs.
+ */
+export function addJavaTypePropertyToSchema(
+  schema_path: string,
+  schema_inst: Record<string, any>,
+  verbose: boolean = false
+) {
+  let java_type_contents = `com.opencaptablecoalition.ocf.schema.${basenameRelativePathToSchemaDir(
```
jacobyavis, line 21: I've added `opencaptablecoalition` as domain here, thanks for the pointer @pjohnmeyer

pjohnmeyer:
> Note that I think $id should be included in raw since it's used in the various $refs

I wouldn't, I'd introduce some kind of short ID to our raw files for refs with some kind of wart like:

`$ref: $enums/EmailType.schema.json`

> it would be nice if the generated schemas which use Composition had the inherited properties exploded with full details

Hmmmm maybe? I'm less sold on this.

I'd be willing to take a stab at this after we get the next release out.

jacobyavis:
> I wouldn't, I'd introduce some kind of short ID to our raw files for refs with some kind of wart like:
> 
> `$ref: $enums/EmailType.schema.json`

Ah yes, duh. I wasn't thinking clearly.

> > it would be nice if the generated schemas which use Composition had the inherited properties exploded with full details

> Hmmmm maybe? I'm less sold on this.

Having these properties exploded would most likely help the schema to behave properly with various JSON Schema tools out there (besides just jsonschema2pojo, which I'm focused on right now), because our composition pattern using `allOf` is non-standard. 

For example, think how much of an easier time we might have had writing the doc generator if the schema properties were exploded. But obviously for development, we want the raw src to be simplified/pure, no chance for unintentional overrides

JSv4:
> > I wouldn't, I'd introduce some kind of short ID to our raw files for refs with some kind of wart like:
> > `$ref: $enums/EmailType.schema.json`
> 
> Ah yes, duh. I wasn't thinking clearly.
> 
> > > it would be nice if the generated schemas which use Composition had the inherited properties exploded with full details
> 
> > Hmmmm maybe? I'm less sold on this.
> 
> Having these properties exploded would most likely help the schema to behave properly with various JSON Schema tools out there (besides just jsonschema2pojo, which I'm focused on right now), because our composition pattern using `allOf` is non-standard.
> 
> For example, think how much of an easier time we might have had writing the doc generator if the schema properties were exploded. But obviously for development, we want the raw src to be simplified/pure, no chance for unintentional overrides

I like this too. I have some python code that does most of this for my own purposes. Happy to contribute or refactor in TS.

jacobyavis:
@JSv4 @pjohnmeyer Can we merge this in the meantime? 
For us to move forward I'd like the javaType to be available at the remote refs by the time we cut the 1.1.0 release

regebro:
I'm not sure why this would be necessary, the information is not actually new information, it's just the location of the schema file, transformed into a Java object name in a simple transformation. That seems to me like it could be a part of the tool that generates the java types.

JSv4:
@jacobyavis, could this just be part of the loader script in whatever tooling you're working on?

jacobyavis:
For sure, it could be just part of the tool that loads the java types. I definitely debated keeping this whole thing from the TWG, but there's two main reasons I decided to create this PR.

1. If there's another corp to come along and decide that jsonschema2pojo is the simplest way forward for them, they will likely run into the same issue/bug that I did. If this change goes in, they would find that integration much more seamless.

2. As I have it locally today, I need processing to change all $refs to local references because the actual remote refs do not carry the `javaType` definition that the lib is looking for

jacobyavis:
After discussion in the TWG meeting yesterday, going to look into creating a separate repo, something like ocf4java which does the required schema manipulations for full jsonschema2pojo compatibility, then deliver the gen'd files as a jar.

PR #426, opened 2023-05-16
Provide for Warrant Issuances with Min, Max, Estimated, Unspecified or Indetermine Quantities

### LINKED ISSUE #372: [Enhancement]: Proposal to Tweak CustomConversionMechanism to Provide for Min/Max

JSv4:
#### What type of PR is this?

/kind bug

As discussed in #372, there is a desire to have some concrete value for warrant conversions in OCX. For modeling purposes a max or an estimated value is OK, so long as it's clear when the quantity is not fixed in the instrument itself. We will support Warrants missing an estimate or known value in OCX, however, as not all platforms collect estimated values and not all warrants have fixed quantities. They'll go in a table where you can provide your own estimates within OCX for modeling purposes.

This PR makes a few changes to address these two issues:
- `exercise_price` is no longer required on warrant issuances
- `quantity` is no longer required on warrant issuances
- new `quantity_source` property on Warrant Issuance event allows you to specify where the quantity came from using `QuantitySourceType` enum. For now, it's implicitly UNSPECIFIED where quantity is provided but quantity_source is not. **I'd like to require quantity_source where quantity is provided in the future (v2.0.0).**

Fixes #372


schema/types/QuantitySourceType.schema.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/enums/QuantitySourceType.schema.json",
+  "title": "Enum - Quantity Source Type",
+  "description": "Enumeration of quantity source types describing where a quantity value came from",
+  "type": "string",
+  "enum": [
+    "USER_ESTIMATED",
```
pjohnmeyer, line 8: Not sure I like this naming (yes I know I proposed it) because I don't know that it's clear who the "USER" is in different processing contexts.

JSv4: Updated in latest commit. Thx for the suggestion.

JSv4: Updated and ready for review.

jacobyavis (approved):

jacobyavis:
Approved, @JSv4 
I'd recommend we start that WIKI for tracking changes we want to make in 2.0, before we lose track.
There's 3 already I can think of:
* this here, making quantity_source required where quantity is provdided
* the big transaction changes, proposed in the PR around RSA issuance
* the small boolean item from earlier today (early exercisable) wherein we decided we could be using `$comment` to also help us find these quickly later

JSv4:
Thanks, @jacobyavis. Good idea on updating the wiki - [updated](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/wiki/Planned-2.0-Rework#deprecation-tracking-tooling)

PR #427, opened 2023-05-16
Add Valuation Based Conversion Mechanism for Warrants

### LINKED ISSUE #362: [Enhancement]: Enhance Warrant Conversion Spec

JSv4:
#### What type of PR is this?

/kind feature

As discussed in https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/issues/362 (and related to the points raised above), Warrants often don't have a fixed exercise price. Related to the above, the quantity of resulting shares may not be fixed and may be dependent on the company's valuation and/or capitalization like SAFEs. Our current warrant issuance requires an exercise price, however, and doesn't support valuation-based exercise calculations like SAFEs. Split these changes out from PR #410.

There is now a `VALUATION_BASED_CONVERSION_MECHANISM` which has a number of the fields from the `SAFE_CONVERSION_MECHANISM` related to the value of the company / size of the round. This `VALUATION_BASED_CONVERSION_MECHANISM` is now permitted for warrants. Could make sense to make it available on other instruments too.

In reviewing some of the sample language, I realized we also wanted a price-per-share based conversion mechanism where we have a conversion amount determined by some discount from a future round's price. Basically, we take the future round price - some discount and divided the original purchase price of the warrant (or security) by that amount. `PricePerShareBasedConversionMechanism` is currently only available for warrants but seems like there's no reason we couldn't make it available for other equity types.

Fixes #362


jacobyavis (approved):
Looks good John, it's past time we get this change in!

PR #456, opened 2023-11-21
fix: Updated quantity source type to live under the enums package instead of the types package

lavens:
#### What type of PR is this?
/kind bug

This PR moves `QuantitySourceType.schema.json` to the `enums` package to match the path defined in the `WarrantIssuance.schema.json`. This change addresses an issue encountered when using a JSON Schema code generation tool.

Separately, the QuantitySourceType is keyed as `"quantity_source"` where enums are typically postfixed with `type`. Looking for guidance on if `quantity_source` should be updated to `quantity_source_type` for consistency, and what the process is for handling a breaking change.


JSv4:
@lavens, sorry we missed this last year. I like the suggestion from the Mantle team to have a response bot that can give people a clearer expectation of response time and process. We'll take a look at this. In order to merge it needs a format review in GitHub and typically, if it's a substantive change, we'd discuss and approve it at the TWG level. For a bug fix, a code review and approval would be sufficient.

schema/enums/QuantitySourceType.schema.json
```diff
@@ -12,5 +12,5 @@
     "INSTRUMENT_MAX",
     "INSTRUMENT_MIN"
   ],
-  "$comment": "Copyright © 2023 Open Cap Table Coalition (https://opencaptablecoalition.com) / Original File: https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/tree/main/schema/types/QuantitySourceType.schema.json"
+  "$comment": "Copyright © 2023 Open Cap Table Coalition (https://opencaptablecoalition.com) / Original File: https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/tree/main/schema/enums/quantitysourcetype.schema.json"
```
JSv4, line 15: This schema is actually in the wrong place too... `https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/main/schema/types/QuantitySourceType.schema.json`

It needs to be moved into the enums folder too or various tools wouldn't work properly.

JSv4, line 15: We need to make changes where this is referenced too:

https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/fec5c3962e15735caf0607b48d3ae2484efea59e/schema/objects/transactions/issuance/WarrantIssuance.schema.json#L72-L77

JSv4: We would leave this in camel case too to conform with other schemas.

JSv4: Oh, just realizing this is just in the wrong place. The $id is perfectly fine and doesn't need to be changed.

JSv4 (changes requested):
Thanks for this @lavens. Looks like this is just in the wrong place, though the `$id` is correct and references the `enum`s folder despite the file being in `types`, thankfully. The `$comment` tag is incorrect due to it being automatically generated from the relative path to the repo root. You don't need to edit it yourself and it should be automatically generated by pre-commit hooks (it should be CamelCase). If you `npm install` from the repo root, these should run automatically when you create a commit. You also have access to a bunch of convenience scripts registered in `package.json`:

https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/fec5c3962e15735caf0607b48d3ae2484efea59e/package.json#L5-L16

JSv4 (approved):
LGTM! Thx

PR #467, opened 2024-02-15
removes home links from all pages

osis:
#### What type of PR is this?

/kind documentation

This removes the home links from all pages. I was initially going to make a PR to more consistently applied the placement of the home link. After browsing through the site for awhile I noticed the navigation is always visible, and the nav hierarchy of the site is very flat, it feels like it might not be needed since that UX pattern is usually for getting a user out of a navigation rabbit hole.


JSv4:
Thanks @osis. No qualms removing this. Just for some background, it predates us having a documentation site with a nice nav, when we needed more navigation aid inside of the GitHub-repo-rendered docs. Do have one minor comment. See my review.

JSv4 (changes requested):
Thanks for doing this. FYI, we came up with a workaround to be able to rename `PlanSecurity` to `EquityCompensation` without introducing breaking changes. It involved using alias object IDs for the old object types. We also left up our documentation for various `PlanSecurity...` events, though those will be deprecated in v2 (whenever that is). 

If you search for ":house: [Documentation Home]", you'll see there are a number of PlanSecurity... related files with the old header still:

image

Can you make sure those are cleaned up too?

osis:
Thanks for the review! Made the changes.

JSv4 (approved):

PR #468, opened 2024-02-26
Added Financing transaction object that can be referenced by Issuance transactions

### LINKED ISSUE #445: [Enhancement]: Tag Specific Issuances with "Financing" Object

In OCX and other lawyer-facing formats, being able to see all of the issuances created as part of a "financing" - however that is defined - is very useful. Adding a financing object and allowing issuances to refer to the financing id would facilitate this.

### Why is this Needed?

Support of human workflows.

Question is whether any platform actually captures this information?

lavens:
#### What type of PR is this?

/kind feature

This PR defines a `Financing` transaction object and allows issuances to refer to it via ID. Per issue #445, this PR would assist in viewing all of the issuances created as part of a "financing".

Fixes #445


JSv4:
Thanks, @lavens. This was a slightly different approach from what I personally envisioned, but not wrong by any means. Can you walk us through it tomorrow? I'd like the group's input. The other approach I think one could take is have financing be an object (but not a `TX_`) and then have the issuance link back to the relevant financing obj. The actual schemas would look pretty similar but there'd be a semantic distinction as to whether a financing is an event or an object. Can see arguments for both.

lavens:
Closed this PR and reopened as https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/pull/478

Issue #457, opened 2023-11-22
[Bug]: Fixing Issuer Authorized Shares Adjustment Validation


victormimo:
### Current Behavior:

1. `ObjectType.schema.json` is missing the enum `TX_ISSUER_AUTHORIZED_SHARES_ADJUSTMENT`, resulting in a failure when validating sample data.
2. `IssuerTransaction.schema.json` references required field `stock_class_id` when it's not in the properties, I believe it was meant to include `issuer_id`

### Expected Behavior:

1. Include the missing enum
2. Change required field to `issuer_id`

### Steps to Reproduce:

1. Add a TX_ISSUER_AUTHORIZED_SHARES_ADJUSTMENT to the sample transaction data
2. Validate it

### Anything else we need to know?

@JSv4 if you give me write access to the repo, I'm happy to create a PR to make reviewing easier.

victormimo:
Will do.

PR #469, opened 2024-02-26
fixing bug related to Issue #457

sachin-shrestha:
What type of PR is this?
/kind bug

ObjectType.schema.json is missing the enum TX_ISSUER_AUTHORIZED_SHARES_ADJUSTMENT, resulting in a failure when validating sample data.
IssuerTransaction.schema.json references required field stock_class_id when it's not in the properties, should be issuer_id


sachin-shrestha:
@JSv4 or @pjohnmeyer ...can one of you review please?

JSv4:
On it @sachin-shrestha

JSv4:
Thanks, @sachin-shrestha, this all looks good to me. Is this something you need merged? If not, I'd like to revisit our versioning tomorrow. We wanted to more formally document that for a while. Since this would technically be a breaking change - even though it fixes a bug - I want to make sure we either (a) version up as appropriate or (b) have a sane and documented policy for the versioning. @pjohnmeyer, would love your thoughts on this. 

FYI, I also opened a new issue #470 to add some structural validation checks to catch both of the issues you fixed here. There's no reason they should have made it through the test suite.

JSv4:
Let's also have a sample here too in the sample transactions file.

jacobyavis:
Added samples, and the description looked okay but I made a minor change

pjohnmeyer:
> Since this would technically be a breaking change - even though it fixes a bug

Just noting for posterity that this was not a breaking change because this object did not exist in v1.1.0.

PR #472, opened 2024-02-26
Expand the Projects List

JSv4:
#### What type of PR is this?

/kind documentation

We've seen a number of new OCF-based open source projects come onto the scene. Updated the project list with the ones I can recall from memory.

N/A

### DIFF
--- a/docs/PROJECTS.md
+++ b/docs/PROJECTS.md
@@ -4,9 +4,17 @@ Publicly available projects or initiatives that work with OCF can be listed here
 
 ## Existing Projects
 
-### Contract Express Exporter
+### CE2OCF Contract Express Exporter
 
-Gunderson Dettmer and @JSv4 are working on OCF export functionality for Incorporation Documents managed in [Contract Express](https://www.thomsonreuters.ca/en/contract-express.html.html), a document and contract automation solution from Thomson Reuters.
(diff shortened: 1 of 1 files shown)


jacobyavis (approved):

jacobyavis:
Hi John, I added ocf4j -- though I just noticed I had never pushed the majority of my work from last spring. I'll attempt to recover and improve that repo in the coming weeks

docs/PROJECTS.md
```diff
@@ -16,6 +24,18 @@ Gunderson Dettmer and @JSv4 are working on OCF export functionality for Incorpor
 
 Shoobx and @regebro are working on PyOCF, a Python library to read and create Open Cap Table Format files. Check it out [here](https://github.com/Shoobx/pyocf).
 
+### OpenCap.co
+
+[OpenCap.co](https://opencap.co/) is both an open source project a company. They're building an [open source cap table application](https://github.com/opencapco/opencap.co) that is compatible with the OCF standard.
```
pjohnmeyer, line 29:
```suggestion
[OpenCap.co](https://opencap.co/) is both an open source project and a company. They're building an [open source cap table application](https://github.com/opencapco/opencap.co) that "will follow" the OCF standard.
```

I don't feel we should claim compatibility on their behalf; "will follow" is what their README says.

jacobyavis: Pushed this change, but I left off the double quotes

jacobyavis:
> Hi John, I added ocf4j -- though I just noticed I had never pushed the majority of my work from last spring. I'll attempt to recover and improve that repo in the coming weeks

I take it back -- removed from the list for now as I recall better what this was about. It's not in a state where it's ready to share, but if/when we do get it there, I will add it back.

PR #474, opened 2024-02-27
commits a formatting diff coming from the hooks

osis:
#### What type of PR is this?

/kind cleanup

Noticing that these files get reformatted when running the pre-commit hooks without any other changes to the codebase. It's very distracting when trying to get other changes in.

I confirmed that my IDE isn't getting in the way and doing some reformatting on its own, but would probably want as least one other person to confirm this is happening as well.


osis:
after going down a bit of a rabbithole, this is due to the [lint-staged cmd](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/main/package.json#L40) being broken in the pre-commit hook. Will submit a fix.

pjohnmeyer (commented):
> Noticing that these files get reformatted when running the pre-commit hooks

Can you explain what "these files" means, and what problem this is fixing? 

I ask because adding the `.` here is not typical for `lint-staged`. `lint-staged` by design passes _only the staged files_ to the command; adding the `.` means prettier is going to format all files, staged or not.

osis:
Hmmmm interesting. `lint-staged` wasn't working for me regardless. I'll return the PR back to it's original state so you can see for yourself.

pjohnmeyer:
> Hmmmm interesting. `lint-staged` wasn't working for me regardless. I'll return the PR back to it's original state so you can see for yourself.

This is probably on me, but I read this as a different problem than what was originally stated, and doesn't answer [my question](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/pull/474#pullrequestreview-1904932651).

Can you please write a "Bug" issue and fill out that template so I know what we're trying to fix?

osis:
Sorry for the lack of clarity @pjohnmeyer. "These files" refers to the files affected in this PR. More details is #476 as requested.

Issue #461, opened 2024-02-04
Evaluate Need for One-to-Many Association between Stock class and Stock plan?


dahal:
First of all, a great initiative. We are currently building OpenSource captable management platform [OpenCap](https://github.com/opencapco/opencap.co) and we will follow the [Open Cap Table Coalition](https://www.opencaptablecoalition.com/format) format as much as possible and would love to collaborate.

I wanted to clarify what the expected associations are between these two tables/schemas, one to one or one to many? cc @JSv4

https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/fec5c3962e15735caf0607b48d3ae2484efea59e/samples/StockPlans.ocf.json#L11

We are going with one to many association, that seems to make more sense.
https://github.com/opencapco/opencap.co/pull/93/files#diff-5b443964f4f3a611682db8f7e02177b0a8c632b2039e2bd5e4dd7347815c565cR165-R217

JSv4:
Hey @dahal, thanks for the question. Each stock plan can only have one stock class linked to it, but you can have multiple plans linked to the same stock class - e.g. you could have three different stock plans all issuing common stock, which would all have the same stock class id. That in line with what you guys are expecting?

jarrettq:
👋 additional context - Canadian corporations do have the ability to designate multiple share "classes" within a given stock/equity plan, which current versions of the OCF spec don't support. (e.g. An option pool of 5,000,000 shares, which can be issued in the form of option grants for either Voting or Non-Voting Common shares).

The Mantle team is going to look to provide some recommendations on how this scenario could be captured with adjustments to the OCF spec.

To make sure we're being as specific as we can be with the wording involved here - with Canadian capitalization definitions, the typical language we could see would indicate:
> (i) an unlimited number of Common Shares, **_issuable in series_**, of which an unlimited number are designated as Voting Common Shares (the “Voting Common Shares”) and an unlimited number are designated as Non-Voting Common Shares (the “Non-Voting Common Shares” and together with the Voting Common Shares, the “Common Shares”) 

So there's an implied hierarchical relationship between the overarching Common Share **class** and the Voting Common & Non-Voting Common **series** within it.

This also winds up being expressed in the Definitions section of Equity Incentive Plans:
> "Common Shares" means the Voting Common Shares or the Non-Voting Common Shares in the capital of the Company;

At which point the individual option grants will specify which series the options would exercise into.

[...]

PR #477, opened 2024-03-12
feat: Update stock plan to support multiple stock class ids

lavens:
What type of PR is this?
/kind feature

This PR adds support for multiple stock classes to be associated with a stock plan.

Which issue(s) this PR fixes:
Fixes https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/issues/461

cc @jarrettq


jarrettq:
Our approach placed an emphasis on not introducing breaking changes, so the following items are worth pointing out:
1. We've opted to mark fields like `stock_class_id` in the Stock Plan object as deprecated, with a goal to eventually sunset it in favor of a `stock_class_ids` list.
2.  `stock_class_id` is still an optional field in the Equity Compensation Issuance transaction, but we'd recommend eventually transitioning it to a required field to handle scenarios where a Stock Plan starts off with a single share class but adds support for other share classes in subsequent updates to it.

schema/objects/StockPlan.schema.json
```diff
@@ -34,13 +34,21 @@
...
+      "items": {
+        "type": "string"
+      }
     },
     "id": {},
     "comments": {}
   },
   "additionalProperties": false,
-  "required": ["plan_name", "stock_class_id", "initial_shares_reserved"],
+  "required": ["plan_name", "stock_class_ids", "initial_shares_reserved"],
```
pjohnmeyer, line 52: Making `stock_class_id` optional is fine, but requiring `stock_class_ids` is a breaking change. We'll need to use a `oneOf` construct to require one or the other within 1.x

@arthur-clara validation thought: As we introduce things like this with deprecation notes, it would be awesome to include deprecation warnings in validator output.

JSv4: @lavens and @jarrettq, here's an example of what @pjohnmeyer is referring to:

https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/7abfda1d5332e816c09b387db95a5f57afbafbb7/schema/objects/transactions/issuance/EquityCompensationIssuance.schema.json#L151-L158

JSv4: @pjohnmeyer, it'd also be great to pick up deprecation notices in a central doc and also give them more prominent billing in the schema-specific docs

lavens: > Making `stock_class_id` optional is fine, but requiring `stock_class_ids` is a breaking change. We'll need to use a `oneOf` construct to require one or the other within 1.x

Good catch, made the update as requested

schema/objects/StockPlan.schema.json
```diff
@@ -34,13 +34,37 @@
...
-      "description": "Identifier of the StockClass object this plan is composed of",
-      "type": "string"
+      "description": "[DEPRECATED in favor of stock_class_ids] Identifier of the StockClass object this plan is composed of.",
+      "type": "string",
+      "deprecated": "true"
+    },
+    "stock_class_ids": {
+      "description": "Identifiers of StockClass objects this plan is composed of",
+      "type": "array",
+      "items": {
```
pjohnmeyer, line 44: Do we need to require the array to contain at least one ID, or is it allowed to be empty?

```suggestion
      "type": "array",
      "minItems": 1,
      "items": {
```

lavens: Yes, the array should contain at least one ID when stock_class_ids is provided

pjohnmeyer (approved):

pjohnmeyer (approved):

Issue #470, opened 2024-02-26
[Enhancement]: Add Structural Validation for Required Field & Object Types Enum


JSv4:
### Description of Enhancement :

As flagged in #457, we had a transaction schema committed for Issuer Authorized Share Adjustment that had the wrong required field name AND an object_type that was not in the object type enum. Either the test failed to trip the GitHub action or our test doesn't cover either situation. I know it doesn't cover checks that fields marked as required actually exist, but I am surprised the missing ObjectType didn't cause a validation error somewhere. 

I propose adding a validation test for BOTH in the validate schema tooling.  

### Why is this Needed?

We didn't catch a schema with two issues making it to a release. 

### Anything else we need to know?

Nope

JSv4:
- This may be as simple as just making sure there is a sample. Can check for that. 
- **ALSO** add additional documentation to walk someone through adding a schema and what is required - e.g. a sample.

jacobyavis:
Solution for this is two-phase

1. Ensure when we're validating schemas that any `object_type`'s `const` value also appears in the enum list on `ObjectType.schema.json`
2. Ensure that for every object type, at least one sample exists in the samples directory

PR #478, opened 2024-03-12
feat: Added Financing object that can be referenced by Issuance trans…

lavens:
What type of PR is this?
/kind feature

This PR defines a Financing object and allows issuances to refer to it via the financing object's ID. Per issue https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/issues/445, this PR would assist in viewing all of the issuances created as part of a "financing".

Which issue(s) this PR fixes:
Fixes https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/issues/445


JSv4:
Thanks @lavens. This looks great. Let's discuss on the call, but this works for me!

pjohnmeyer (commented):
This makes sense to me @lavens, I just want to propose one alternative. Instead of each Issuance referencing a `financing_id`, would it be "better" to have the financing object reference an array of `resulting_security_ids`? Since this field is only interesting if there is a Financing object present, then it seems reasonable to me that we might stuff that information on the financing object.

On the other hand, your approach feels a little more compatible with an API that might be linking or unlinking individual instances.

lavens:
> This makes sense to me @lavens, I just want to propose one alternative. Instead of each Issuance referencing a `financing_id`, would it be "better" to have the financing object reference an array of `resulting_security_ids`? Since this field is only interesting if there is a Financing object present, then it seems reasonable to me that we might stuff that information on the financing object.
> 
> On the other hand, your approach feels a little more compatible with an API that might be linking or unlinking individual issuances?

@pjohnmeyer That makes sense to me. This suggested approach would remove the need to update the issuance object and keep all relevant information isolated to the financing object. I'll make this change as a second commit in this PR so both options can be reviewed by the group.

pjohnmeyer (approved):
Per TWG Call: Agreed to this approach

schema/objects/Financing.schema.json
```diff
@@ -0,0 +1,34 @@
...
+      "type": "array",
+      "items": {
+        "type": "string",
+        "minItems": 1
+      }
+    },
+    "comments": {}
+  },
+  "additionalProperties": false,
+  "required": ["name", "issuance_ids"],
```
jacobyavis, line 32: Should there be a `date` associated with each financing?

jacobyavis:
As a clean-up item, we should update some of the markdown documentation regarding file types
e.g. Architecture.md mentions and lists "eight file types"

Also, we need to add the new file type to `UriLookupForFileType.json` so that the validator utils will consider this new file type.

schema/objects/Financing.schema.json
jacobyavis: The sample file mentions the date in the comments section
```
{
  "file_type": "OCF_FINANCINGS_FILE",
  "items": [
    {
      "object_type": "FINANCING",
      "id": "5bfe12c6-f049-4216-9f94-5a75dc3aaee5",
      "name": "Series A Funding",
      "issuance_ids": ["test-stock-issuance-minimal"],
      "comments": ["A funding round led by Firm Y on 02/15/2024."]
    }
  ]
}
```
In my opinion this data should be structured

jacobyavis:
One more note: we missed adding FINANCING to the ObjectType schema, which was caught by the new validation I', wrking on for #470

https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/actions/runs/8837798105/job/24267465832?pr=491

lavens:
@jacobyavis thanks for the review and suggestions. I did see that a few of the issues  are addressed in the validation PR #491 . That said I created a PR, to isolate the financing related changes, to address the found issues, update the generated schema docs, and include the proposed `date` property, #494

PR #479, opened 2024-03-26
Removed maxLength limit for subdivision and updated description

### LINKED ISSUE #442: [Enhancement]: Remove 2-digit length validation on CountrySubdivionCode

Change the regex pattern to remove the digit length restriction or increase the length to at least 4 digits. (So, allow entries between 2 and 4 digits inclusive: ^[A-Z0-9]{1,5}$
### Why is this Needed?

Right now, the countrySubdivisionCode variable is a string that must match regex pattern: ^[A-Z0-9]{1,3}$

This works for US ISO 3166-2 subdivisions (a.k.a. States), but it does not work in countries where the jurisdiction does not match the US ISO 3166-2 codes. 

For example, two key startup jurisdictions in the UAE are: "Abu Dhabi Global Market" (ADGM) and "Dubai International Financial Centre" (DIFC). It is impossible to put these jurisdictions into 2 digit format.

If we want to keep countrySubdivisionCode as 2 digits to validate properly for the US, then we could add an additional optional variable to the Issuer object called "jurisdiction" or something like that.

arthur-clara:
#### What type of PR is this?
/kind feature

Updates the subdivision schema to remove the limitation to match ISO 3166-2 to allow subdivisions with more than two digits.

Fixes #442


JSv4 (approved):
LGTM!

schema/types/CountrySubdivisionCode.schema.json
```diff
@@ -2,10 +2,9 @@
   "$schema": "http://json-schema.org/draft-07/schema",
   "$id": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/types/CountrySubdivisionCode.schema.json",
   "title": "Type - Country Subdivision Code",
-  "description": "Type representation of the second part of an ISO 3166-2 country subdivision code",
+  "description": "State, province, or equivalent identifier required for an address in this country",
   "type": "string",
   "minLength": 1,
-  "maxLength": 3,
-  "pattern": "^[A-Z0-9]{1,3}$",
+  "pattern": "^[A-Z0-9]{1,}$",
```
pjohnmeyer, line 8: @arthur-clara @JSv4 should we eliminate the pattern entirely? Now that this isn't ISO, who is to say that the only characters are latin alphabet A-Z and digits 0-9

JSv4:
For some context, the ISO codes don't map cleanly to specific jurisdictions that relevant for the purposes of startup financings - e.g. Abu Dhabi

JSv4:
We're thinking that we need to separately handle subdivision name and subdivision code or some kind of split to handle ISO codes where they're of sufficient fidelity but capture finer grains where we need to.

JSv4:
Per discussion on 6/20/24, we can use two properties here - one for codes (for US states) and one for name (where ISO codes don't make sense).

JSv4:
@arthur-clara to add a `oneOf` validation

JSv4:
Makes sense. I say we call this. Becoming the longest little quick PR in all the known universe :-).

arthur-clara:
Can I just add the country_subdivision_name as an optional to the address schema so that we can accomodate non-US?

lavens:
@arthur-clara this might be an approach to enforce a oneof for multiple optional values
https://stackoverflow.com/questions/60403128/json-schema-oneof-without-required

MattCantor (commented):
lgtm

MattCantor (approved):

PR #471, opened 2024-02-26
Add Option Repricing Event

JSv4:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

As discussed in #460, we need a way to reprice outstanding equity compensation (options). This adds a transaction to reprice specific options.  

#### Which issue(s) this PR fixes:

Fixes #460

PR #480, opened 2024-03-26
Creation of consolidation transaction for stock

### LINKED ISSUE #449: Model a Consolidation Transaction for Inclusion to OCF
### Discussed in https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/discussions/444

<sup>Originally posted by **victormimo** August  7, 2023</sup>
Let's assume the scenario below.

Screenshot 2023-08-07 at 5 37 48 PM

Let's also assume the issuance objects for Bob are the same outside of quantity.

I understand that if Bob wants to transfer Alice 1100 shares, then we need to create 2 separate transfers (because transfers only have one security_id field corresponding to Bobs). One for 1000 shares and one for 100 shares, this also means that Alice will have two new Issuance events, one for 1000 and one for 100.

am I right on this logic?

Could there be a way to unify Alice's position into one for 1100 shares? 

If we take a scenario where an issuer issues 1,000,000 shares to a stakeholder, it's likely that a new buyer purchases less than a million shares. This means overtime that issuance of a million shares will result in many smaller issuances since there's no way to unify positions on transfers. Even if someone buys one million shares for the exact `stock_class_id`, they will have to carry however many issuance objects there were to sum to a million.

One potential solution could be to change the transfer's `security_id` field to be an array, like `resulting_security_ids`
Something like.

```
{
      "object_type": "TX_STOCK_TRANSFER",
      "id": "bob-transfer-to-alice",
      "security_ids": ["bob-stock-1", "bob-stock-2"],
      "date": "2023-02-01",
      "resulting_security_ids": [
        "alice-stock-1"
      ],
      "quantity": "1100"
}
```

Let me know if I'm missing something

</div>

arthur-clara:
#### What type of PR is this?
/kind feature

This is the first cut of the addition of the stock consolidation. For review and discussion.

This PR creates a new Stock Transaction called "Consolidation".
See the discussion by VictorMimo on issue 449 on why it is needed.

Fixes #449


schema/primitives/objects/transactions/consolidation/Consolidation.schema.json
```diff
@@ -0,0 +1,29 @@
...
+  "properties": {
+    "security_ids": {
+      "title": "Security Consolidation - Incoming Securities ID Array",
+      "description": "Identifier for the securities (stock, plan security, warrant, or convertible) that are being consolidated into one position by which it can be referenced by other transaction objects. Note that while this identifier is created with an issuance object, it should be different than the issuance object's `id` field which identifies the issuance transaction object itself. All future transactions on the security (e.g. acceptance, transfer, cancel, etc.) must reference this `security_id` to qualify which security the transaction applies to.",
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "resulting_security_id": {
```
JSv4, line 16: @arthur-clara, can you use a minItems validation here to ensure there are two or more ids? See here for an example: https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/7abfda1d5332e816c09b387db95a5f57afbafbb7/schema/primitives/objects/transactions/transfer/Transfer.schema.json#L16-L24

JSv4 (changes requested):
Looks good, @arthur-clara! Couple minor things:
1. Can you merge the latest changes from main back in? I just merged a bunch of PRs and there are a few conflicts and a change to the doc generator which will change the headers in your docs in this PR. 
2. I'd suggest doing minItems: 2 and uniqueItems: True attrs on the Consolidation Transaction primitive.

schema/primitives/objects/transactions/consolidation/Consolidation.schema.json
JSv4: Actually, thinking about #471, @arthur-clara, maybe we can make this more general purpose? Like a multiple security transaction primitive which you can compose into your consolidation transaction. See more here - https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/pull/471#issuecomment-1967283436

JSv4: This should be "MultipleSecurityTransaction" or something similar.

PR #486, opened 2024-04-16
proposed balance-cancellation schema for transfers

### LINKED ISSUE #483: [Enhancement]: Add support for 'balance cancellation' when doing a Transfer

The schema for Stock Transfer does not support a situation where there's a partial transfer with the _balance cancelled_. Currently the only way to do this would be to issue a balance certificate then subsequently cancel it. The problem with this is platforms that support this will not have a `security_id` to reference for the temp balance certificate.

### Why is this Needed?

To properly represent a partial transfer+ balance cancellation transaction (such as is possible in Shareworks), since TX_STOCK_TRANSFER expects all shares to go to new security_ids (either transferee certificates or a balance certificate)

jacobyavis:
#### What type of PR is this?
/kind feature

To discuss: when is a transfer considered to be a terminal transaction for a given security_id
With this proposed boolean flag on the transfer objects, OCF can better determine validity of transfer scenarios

Fixes #483


schema/primitives/objects/transactions/transfer/Transfer.schema.json
```diff
@@ -22,6 +22,11 @@
       },
       "minItems": 1,
       "uniqueItems": true
+    },
+    "is_balance_cancelled": {
+      "type": "boolean",
+      "description": "If this is a partial transfer, is the balance cancelled? If yes, we expect no balance_security_id here, and we expect this security_id to also be referenced in a cancellation transaction.",
+      "$comment": "REQUIRED in v2"
```
jacobyavis, line 29: copied this $comment style that @JSv4 introduced in PR #428

PR #487, opened 2024-04-17
Add document describing our versioning policy

### LINKED ISSUE #416: Create document describing our versioning policy
As discussed at multiple TWGs and in issue discussions such as [this one](https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/pull/397#issuecomment-1521069081), we would like to explicitly define our versioning policy. This would be based on [semantic versioning](https://semver.org/spec/v2.0.0.html). We do, however, want to create a distinction between the "interface" and "implementation" parts of our schema, so that we have the flexibility to change folder structure, layouts, general approaches, etc., while maintaining stability. The proposal on the table is to: maintain all `files` and `objects` $ids through the duration of 1.0; users of the schema are "allowed" to reference those IDs directly and expect the only thing to change is the version number. Anybody directly referencing the $id of a schema under `enums`, `primitives`, or `types` is bypassing the "interface" of the schema and is subject to potential breakage.

pjohnmeyer:
#### What type of PR is this?
/kind documentation

This is the proposed explanation of how semantic versioning applies to our JSON Schema, so folks contributing to the schema know what they may and may not "break".

The `docs/development` folder will be used to add more documentation as we update our CONTRIBUTING guidelines.

Resolves #416

### DIFF
--- a/docs/development/versioning.md
+++ b/docs/development/versioning.md
@@ -74,9 +74,9 @@ In this case, we maintain backwards compatibility by:
 3. Use `oneOf` to require _either_ the old property _or_ the new one, to prevent
    ambiguity. Example:
 
-    "oneOf": [ { "required": ["old_property"], "not": { "required":
-    ["new_property"] },
-    "$comment": "Due to how the JSONSchema 'not' works, this means that, if old_property is present, new_property cannot be present"
+   "oneOf": [ { "required": ["old_property"], "not": { "required":
+   ["new_property"] },
+   "$comment": "Due to how the JSONSchema 'not' works, this means that, if old_property is present, new_property cannot be present"
         },
         {
         "required": ["new_property"],
@@ -84,8 +84,8 @@ In this case, we maintain backwards compatibility by:
             "required": ["old_property"]
         },
         "$comment":
-    "Due to how the JSONSchema 'not' works, this means that, if new_property is
-    present, old_property cannot be present" } ],
+   "Due to how the JSONSchema 'not' works, this means that, if new_property is
+   present, old_property cannot be present" } ],
 
 ### Renaming a Schema in the Public API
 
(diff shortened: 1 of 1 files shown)


jacobyavis:
Great doc, Patrick. Looks good to me!

jacobyavis (approved):

jacobyavis (approved):

docs/development/versioning.md
```diff
@@ -0,0 +1,107 @@
...
+> **Exception:** A pull request submitted to _fix_ a breaking change that was mistakenly introduced
+> is allowed to be a "breaking" change when compared to the flawed releases, without incrementing
+> the MAJOR version number, as long as it maintains compatibility with older releases. The faulty
+> releases will be deprecated and/or unpublished.
+
+### Renaming a Required Property
+
+If we determine a required property needs to be renamed, we still have to support the old property
+name until a new MAJOR version of the schema is released. One specific reason we have renamed a
+property was to support multiple values where only one value was previously supported.
```
lavens, line 62: @pjohnmeyer Any thoughts on how a replaced field should be tagged as "old"/"deprecated" so that it's clear what needs to be cleaned up when a new MAJOR version of the schema is released? I've come across a few instances where deprecated values are called out in schema descriptions. If that works, it could be helpful to formalize that process here as well.
eg `"description": "This is done to avoid a breaking change as we work towards a bigger restructure of the equity types in v2.0.0. `TX_PLAN_SECURITY_ACCEPTANCE` will be deprecated in v2.0.0"`

lavens (approved):
LGTM! One minor question regarding tracking "deprecated" properties

docs/development/versioning.md
jacobyavis: In the Wiki tab in GitHub, John had started a document called "Planned 2.0 Rework" with a running list of breaking changes. There's also a note about how we started using `$comment` to denote some v2 changes

pjohnmeyer: `$comment` works for some things but not for others -- but it could definitely work for a replaced field, as seen in:

https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/blob/24199843757bb78db87ad90e50d7a7f98d4ae37f/schema/objects/transactions/issuance/EquityCompensationIssuance.schema.json#L38-L42

I wonder, can we use the `deprecated` keyword as [defined in JSON Schema 2019-09](https://json-schema.org/draft/2019-09/json-schema-validation#rfc.section.9.3) even though we're currently on draft-7? Will that break existing tooling, or will it just sit passively until we move to a newer version of JSON Schema? If it can sit passively that would actually be my preference.

JSv4:
@lavens, can you make a suggested change here or are we good to merge and come back around with changes?

pjohnmeyer:
@JSv4 @jacobyavis can one of you please either pick this up for me or merge it as is?

jacobyavis:
Going to merge this as is, as it does provide value. We can improve in the future when time permits

PR #497, opened 2024-05-02
add a schema for a Documents file (paths only) to be included in OCF

### LINKED ISSUE #496: [Enhancement]: Adding a Document schema (paths only) to OCF

While we're not sure about the idea of bloating .ocf.zip archives with gigabytes of documents, we DO see value in OCF being able to support a mapping of` document_ids` to document _paths_.

It's been a while since we talked about including documents in OCF, so I thought it'd be a good time to do a proposal.

### Why is this Needed?

There are many OCF use cases wherein a bunch of documents are created related to the cap table data, and the current OCF schema has no way of supporting any sort of document-related data-mapping.

Use case e.g. Issuer exports OCF data from platform to do a tender offer with 3rd party provider, which generates documents. Third party provider could export the data back including document links

In this proposal, transmission of the actual document archive would remain outside the scope of OCF. But we would support OCF to know the paths within said archive as they relate to the individual stakeholders.

We will come up with a PR to better illustrate our proposal.

jacobyavis:
#### Proposal
proposal for a Documents file to be included in OCF to support mapping of documents to stakeholders and their related securities (note: documents themselves still remain outside of OCF, we are just mapping by way of the path/filename)

#### What type of PR is this?
/kind feature

This PR allows us to start relating document (paths/filenames) with stakeholders, and optionally to their related securities.
* We are NOT proposing to start packaging documents inside of an OCF container
* We ARE proposing there is value in being able to specify filepaths and where they relate to, such that an issuer could provide a separate archive/container/file-repository which these filepaths can be used with

Fixes #496


samples/Stakeholders.ocf.json
jacobyavis, line 1: This samples file was severely lacking, so I re-did it, being sure to include both an Individual Stakeholder and an Institutional Stakeholder with all fields

schema/files/OCFManifestFile.schema.json
```diff
@@ -100,7 +98,14 @@
...
+        "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/types/File.schema.json"
+      }
+    },
+    "documents_files": {
+      "title": "Cap Table - Document File(s) Array",
+      "description": "List of files containing lists of document paths, indexed from the file containing the first such object created to the file containing the last (See separate /schema/files/documents_file schema to validate loaded files)",
+      "type": "array",
       "items": {
         "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/types/File.schema.json"
       }
```
jacobyavis, line 111: Note in the `required` array (below the shown diff) we have the original 7 non-manifest files as required.

When `financings_files` was added, we did not require it. Similar we will not require `documents_files` at this time.

I also noticed some files having `"minItems": 1` but not all. Since adding `"minItems": 1` where it doesn't exist here is technically a breaking change, I removed `"minItems": 1` from the couple places it did exist (empty array is fine in this scenario IMO)

samples/Stakeholders.ocf.json
jacobyavis: Side-note: in the Stakeholder object schema, we specify that `contact_info` prop be used with `INDIVIDUAL` stakeholders, and `primary_contact` prop be used with `INSTITUTIONAL`
However, this is not enforced. 
Do we care?

pjohnmeyer:
Per TWG: @jacobyavis to update so that documents reference:

1. An array of OCF objects,
2. Polymorphically by ID+object type

schema/objects/Document.schema.json
```diff
@@ -0,0 +1,30 @@
...
+    },
+    "security_id": {
+      "description": "Identifier for the security, if any, which this document relates to",
+      "type": "string"
+    },
+    "id": {},
+    "comments": {}
+  },
+  "additionalProperties": false,
+  "required": ["path"],
```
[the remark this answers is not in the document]
jacobyavis: I've now updated this to allow any type of object to be related to a document
I introduced new type `ObjectReference` which is a pair of `object_type` and `object_id`

jacobyavis:
There were comments from last meeting about adding a file hash field, as well as a URI field (both optional)

jacobyavis:
Looking for review (@JSv4 or @pjohnmeyer ?) following the implementation of @mrosendin's uri and md5 suggestion

schema/objects/Document.schema.json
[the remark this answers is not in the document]
JSv4: Couple questions: 
1. Do we ever want/anticipate the file path being a uri? What if we had a static, public url hosting the file? 
3. Also what if someone doesn't want / can't make an archive public at a URI?

docs/explainers/Architecture.md
```diff
@@ -90,6 +90,8 @@ _There are currently 9 file types that make up a cap table_:
    storing vesting schedules used by the issuer.
 9. [Financings File(s)](../schema_markdown/schema/files/FinancingsFile.md) - One or more files
    storing the issuer's financing events.
+10. [Documents File(s)](../schema_markdown/schema/files/DocumentsFile.md) - One or more files
+    storing references to related stakeholder documents.
```
JSv4, line 94: Let's also add some brief documentation on how the document package would work here. If it's riding alongside the normal ocf file in a separate zip, we should describe this.

jacobyavis: Updated, please review. Thanks!

JSv4 (approved):
Looks awesome!

PR #515, opened 2024-08-15
add schemas to support attaching an exact array of vestings to any issuances with vesting (as an option to replace the need for vesting_terms_id)

### LINKED ISSUE #509: [Enhancement]: Enable a "Manual" Vesting Schedule Attached Directly to Issuances
## Description of Enhancement :

We've been doing more work with export and import, particularly of equity compensation and vesting schedules. Folks are finding some of the most complex issuances are bespoke and require that we generate vesting schedules from them only to then trust that the imported will perfectly import them on the other side. We think it would be safer and easier to allow someone to just create one-off or adhoc schedules with list of dates and vesting amounts that can be attached directly to issuances. 

### Why is this Needed?

Reduce data bloat and cut down on # of times vesting schedules need to be translated from data and amounts to OCF vesting schedule objects. 

Nope

jacobyavis:
#### What type of PR is this?

/kind feature

To allow an implementer of OCF to forego the VestingTerms schema in favour of a simpler "manual" style schedule (a simple list of dates and amounts)

Fixes #509


schema/objects/transactions/issuance/EquityCompensationIssuance.schema.json
```diff
@@ -58,9 +58,18 @@
       "$comment": "REQUIRED in v2"
     },
     "vesting_terms_id": {
-      "description": "Identifier of the VestingTerms to which this security is subject.  If not present, security is fully vested on issuance.",
+      "description": "Identifier of the VestingTerms to which this security is subject. If neither `vesting_terms_id` or `vestings` are present then the security is fully vested on issuance.",
```
JSv4, line 61: Now that we have the validator project, when we write natural language rules like this, let's also open a matching issue in the validator repo and cross-link the issues. That way we can finally start tracking these sorts of things. Even if it takes some time to get the rule into the XState code, we'll be able to keep track.

JSv4:
Thanks for putting this together @jacobyavis

schema/objects/transactions/issuance/EquityCompensationIssuance.schema.json
jacobyavis: Good idea.
opened https://github.com/Open-Cap-Table-Coalition/OCF-XState-Validation/issues/76

JSv4 (approved):
LGTM

PR #516, opened 2024-08-15
449 consolidation transaction

### LINKED ISSUE #449: Model a Consolidation Transaction for Inclusion to OCF

arthur-clara:
#### What type of PR is this?

/kind feature

Add a stock consolidation transaction type

Fixes #449


schema/primitives/objects/transactions/consolidation/Consolidation.schema.json
```diff
@@ -0,0 +1,30 @@
...
+      "description": "Array of identifiers for the security (or securities) being consolidation as a result of the transaction",
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "minItems": 1,
+      "uniqueItems": true
+    },
+    "reason_text": {
+      "title": "Reason for stock reissuance",
```
JSv4, line 23: change to consolidation.

JSv4 (changes requested):
Please correct the `reason_text` - e.g.:

```
"reason_text": {
      "title": "Reason for stock reissuance",   
```

JSv4 (approved):

PR #530, opened 2024-09-12
add schemas to support a Change Event concept within the transactions file

### LINKED ISSUE #373: [Enhancement]: Proposal to Make `current_relationship` on Stakeholder a list

This was discussed quite some time ago (over a year). We debated making relationship a multi-select and decided against it. I'm remembering now that it may have been related to not having the version history for changes to that field and concerns about the list going stale. Think there was also some concern with not all vendors having this information and, if they export a single relationship, it would not be clear whether this is the entirety of the relationships a given stakeholder has or whether the vendor just doesn't track it. Anyway, there is a *lot* of interest in the legal working group having a list for this field, possibly making it a list of objects with start and end dates for the various relationship objs. 

### Why is this Needed?

Lot of interest in having more data in this field, particularly because current approach is known to be unable to capture common situations - e.g. ex-employee who becomes an adviser or someone who is a board member and then becomes an investor. An issue here, again, is all vendors may not have this.

### LINKED ISSUE #443: [Enhancement]:creating a property for stakeholder object: status

Should create a property for a stakeholer object, which is "status", is used to describe the staus of this stakeholder.  for example, active, inactive, terminated, for the purpose to provide a complete and clear idea of a specifi stakeholder.

### LINKED ISSUE #522: [Enhancement]: Support OCX Termination Date Column

Cannot currently support termination date col in OCX. We had explored add and remove relationship events in #373 and also status changes in #443. @jacobyavis to review both and propose Stakeholder events.

jacobyavis:
add schemas to support a Change Event concept within the transactions file, with the first two types being Stakeholder Relationship changes and Stakeholder Status changes. Also includes a current_status field on the Stakeholder object for consistency

#### What type of PR is this?
/kind feature

This PR introduces a set of Stakeholder "Change Events" within the transactions file.

This is similar to the existing transactions we have for Security, Stock Class, Stock Plan, and  Issuer -- though I didn't want to call it a "transaction" in this case because the change does not have anything to do with cap table values changing, as is the case with all of our other transaction types to date

Fixes #373
Fixes #443
Fixes #522


schema/objects/transactions/change_event/StakeholderRelationshipChangeEvent.schema.json
```diff
@@ -0,0 +1,45 @@
...
+      "const": "CE_STAKEHOLDER_RELATIONSHIP"
+    },
+    "relationship_started": {
+      "description": "Denoting the beginning of this relationship on the change date",
+      "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/enums/StakeholderRelationshipType.schema.json"
+    },
+    "relationship_ended": {
+      "description": "Denoting the ending of this relationship on the change date",
+      "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/enums/StakeholderRelationshipType.schema.json"
+    },
```
jacobyavis, line 29: since there can exist multiple relationships at the same time, this change event uses an additive/subtractive approach. We can have a change events that adds a new relationship (via `relationship_started` prop), a change event that removes an old relationship (via `relationship_ended` prop) or a combination of both in the same change event

schema/objects/transactions/change_event/StakeholderStatusChangeEvent.schema.json
```diff
@@ -0,0 +1,34 @@
...
+    }
+  ],
+  "properties": {
+    "object_type": {
+      "const": "CE_STAKEHOLDER_STATUS"
+    },
+    "new_status": {
+      "description": "Denoting the beginning of this activity status on the change date",
+      "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/enums/StakeholderStatusType.schema.json"
+    },
```
jacobyavis, line 25: since there can only be one status for a given stakeholder, this doesn't use the "started" / "ended" approach that we took with relationships, but rather just "new" -- what is the new status after this event?

schema/objects/Stakeholder.schema.json
```diff
@@ -29,6 +29,10 @@
       "description": "What is the current relationship of the stakeholder to the issuer?",
       "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/enums/StakeholderRelationshipType.schema.json"
     },
+    "current_status": {
+      "description": "What is the current activity status of the stakeholder?",
+      "$ref": "https://raw.githubusercontent.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/main/schema/enums/StakeholderStatusType.schema.json"
+    },
```
jacobyavis, line 35: added `current_status` to go alongside `current_relationship`

I thought of maybe changing these to be initial values instead, but I don't think that's as useful at a glance, nor would many platforms be able to provide this reliably

schema/enums/StakeholderStatusType.schema.json
```diff
@@ -0,0 +1,19 @@
...
+    "TERMINATION_VOLUNTARY_OTHER",
+    "TERMINATION_VOLUNTARY_GOOD_CAUSE",
+    "TERMINATION_VOLUNTARY_RETIREMENT",
+    "TERMINATION_INVOLUNTARY_OTHER",
+    "TERMINATION_INVOLUNTARY_DEATH",
+    "TERMINATION_INVOLUNTARY_DISABILITY",
+    "TERMINATION_INVOLUNTARY_WITH_CAUSE"
+  ],
+  "$comment": "Copyright © 2024 Open Cap Table Coalition (https://opencaptablecoalition.com) / Original File: https://github.com/Open-Cap-Table-Coalition/Open-Cap-Format-OCF/tree/main/schema/enums/StakeholderStatusType.schema.json"
+}
```
lavens, line 19: @jacobyavis  Is there logic that we foresee driven by the `StakeholderStatusType` (specifically the various inactive reasons) or do we see this as additional metadata to explain the reason why a stakeholder is active/inactive. If it's the latter, what are your thoughts if an enum with the values of `ACTIVE` and `INACTIVE` and an accompanying `StatusReason` of type string in the `Stakeholder` schema could accomplish the same task while allowing for more flexibility.

lavens: That makes sense to me, thanks for the context!

schema/objects/Stakeholder.schema.json
JSv4: And the idea here @jacobyavis is that we'd update the stakeholder objects on export to always represent the most recent status? 

Since the change events support multiple relationships to the issuer, shouldn't current_relationship be a list?

JSv4 (commented):
Love it, but see my comment re current_relationship.

ShareworksLS (approved):

schema/objects/Stakeholder.schema.json
jacobyavis: Good point. I've now introduced `current_relationships` as an array, and marked the singular version with a deprecation warning

PR #541, opened 2025-01-10
Add IssuerAuthorizedSharesAdjustment to the list of transaction classes

regebro:
Add IssuerAuthorizedSharesAdjustment to the list of transaction classes in TransactionFile

#### What type of PR is this?

/kind bug

PyOCF fails schema validation if you add a IssuerAuthorizedSharesAdjustment to the items in a TransactionFile.
PyOCF v1.2.0b1 uses a fork of this schema with the same change, and then it works.

I didn't file a separate issue, as this change is so small that we can discuss it on the PR.


arthur-clara (approved):
LGTM.

PR #578, opened 2026-05-16
Fix SchemaWriter to empty the actual docs output directory (#577)

### LINKED ISSUE #577: Orphaned doc: docs/schema_markdown/schema/objects/transactions/repricing/EquityCompensationReprice.md
## Problem

`docs/schema_markdown/schema/objects/transactions/repricing/EquityCompensationReprice.md` is committed to `main` but has no corresponding schema and is not produced by `npm run docs:generate`.

The schema on disk is `schema/objects/transactions/repricing/EquityCompensationRepricing.schema.json` (note `Repric**ing**`), so the doc generator emits `EquityCompensationRepricing.md`. The older `EquityCompensationReprice.md` (no `-ing`) has lingered in the committed tree since `ef8b8d69` ("Update the path from .../reprice/... to .../repricING/..."), which renamed the parent directory but did not delete the old filename.

## Reproduction

```bash
rm -rf docs/schema_markdown
npm run docs:generate
git status docs/schema_markdown
# shows:
#   D docs/schema_markdown/.pages
#   D docs/schema_markdown/schema/objects/transactions/repricing/EquityCompensationReprice.md
```

`.pages` is a legitimate static mkdocs file (lives outside `schema_markdown/schema/`, which is the only subtree `SchemaWriter` empties). The repricing file, however, is a stale generator output.

## Why `check-gen-docs` CI hasn't caught it

`SchemaWriter.deleteExistingFiles` only empties `docs/schema_markdown/schema/` and then `writeNewFiles` writes the 175 schema-derived markdown files. Because the writer never deletes files that no longer have a corresponding schema, the orphaned `.md` survives every regeneration and the diff-against-committed CI job stays clean.

## Suggested fix

- Delete `docs/schema_markdown/schema/objects/transactions/repricing/EquityCompensationReprice.md` from the committed tree.
- Optionally tighten `SchemaWriter` so that future renames/removals don't leave orphans: e.g. compute the expected output paths from `schema.schemaNodes` and remove any pre-existing `.md` under `docs/schema_markdown/schema/` that isn't in that set, instead of (or in addition to) calling `emptyDir`. `emptyDir` already removes everything under `schema_markdown/schema/`, so this orphan only persists because it gets re-committed; a stricter version would catch any future drift.

JSv4:
## Summary

- Fixes #577. `SchemaWriter.deleteExistingFiles` was calling `emptyDir(path.join(this.output, "schema_markdown/schema"))`, but the writer outputs every file under `docs/schema_markdown/...`. The missing `docs/` prefix meant `emptyDir` was operating on (and creating!) a stray `schema_markdown/` directory at the repo root, while the real `docs/schema_markdown/schema/` was never cleaned between runs. Schema renames or removals therefore left orphan `.md` files in the committed tree forever (e.g. `EquityCompensationReprice.md` left behind by the `reprice` → `repricing` rename in `ef8b8d69`), and `check-gen-docs` stayed clean because regeneration is idempotent on whatever survives.
- Adds `docs/` to the `emptyDir` path so the schema markdown directory actually gets emptied before each regen, and deletes the existing orphan. The sibling `docs/schema_markdown/.pages` is preserved because it lives outside the `schema/` subtree.
- Adds `SchemaWriter.test.ts` covering three cases that all fail against `main`: writing the expected file, removing a pre-seeded orphan, and not creating a spurious `schema_markdown/` directory at the output root.

## Test plan

- [x] `npm test` (15 suites, 27 tests, all green)
- [x] `npm run lint` (only pre-existing untracked `CLAUDE.md` warns)
- [x] `npm run schema:validate-ocf-file-schemas`
- [x] `npm run schema:validate-example-ocf-files`
- [x] `npm run schema:validate-all-objects-have-samples`
- [x] `npm run docs:generate` twice in a row — second run yields a clean `git status` and no stray `schema_markdown/` at repo root

Closes #577


JSCRUDAT0 (approved):
+1

PR #576, opened 2026-05-16
Extract SchemaComposer utility and add schema:compose CLI

JSv4:
## Summary

- Pulls `allOf`-chain flattening out of the doc-generator's `SchemaNode` classes into a new framework-free module: `utils/schema-utils/SchemaComposer.ts`. Exports `composeSchema`, `composeAll`, `buildSchemaRegistry`, and `isBackwardsCompatibleWrapper`.
- The doc generator's `Schema` constructor now invokes `composeAll(rawJsons, { onMissingRef: "skip" })` before building `SchemaNode`s, so each node sees an already-flattened JSON. `SchemaNode.properties()` and `required()` become straight reads; `allOf()` is retained for "Composed From" link rendering. The wrapper class loses its dead overrides because wrapper-detection now lives in the composer.
- New `npm run schema:compose -- --out <dir>` CLI (`utils/schema-utils/ComposeSchemas.ts`) walks `/schema`, composes every file, and writes the full registry to `<dir>` mirroring the source layout. The target directory (including missing parents) is created on demand.

## Behavior preservation

- `diff -r <pre-change-docs> docs/schema_markdown` is **empty** — every byte of the 176 generated markdown files is identical to `main`. The `check-gen-docs` CI job will pass without any commit to `docs/schema_markdown`.
- Wrapper schemas (e.g. `objects/transactions/cancellation/PlanSecurityCancellation`) pass through the composer untouched, matching the previous renderer's behavior.
- A composed top-level object (e.g. `StockConversion`) now exposes every inherited property at the parent's position, with the child's own values winning on overrides and child-side `{}` placeholders deferring to the parent — exact replication of the prior in-class merge logic.

## Test plan

- [x] `npm test` — 15 suites, 38 tests pass (14 new tests in `utils/schema-utils/SchemaComposer.test.ts` cover wrapper detection, registry building, property/required merging, key-order preservation, placeholder fallback, missing-ref errors, and memoization)
- [x] `npm run docs:generate` — output identical to `main`'s `docs/schema_markdown` (verified with `diff -r`)
- [x] `npm run schema:validate-ocf-file-schemas` — all schemas valid
- [x] `npm run schema:compose -- --out /tmp/ocf-composed` — writes 175 composed schemas (175 source files in `/schema`)
- [x] `npm run schema:compose -- --out /tmp/nested/missing/dir` — creates nested target directory on demand
- [x] `npm run schema:compose -- --help` — prints usage; missing `--out` errors with `Missing required argument: out`

PR #581, opened 2026-06-17
Add schema toolkit: TypeScript codegen and dereferencer

JSCRUDAT0:
Stacked on #576 (`refactor/schema-composer`) — review/merge that first; the base will retarget to `main` once #576 lands.

Adds two framework-free utilities on top of the `SchemaComposer`, sharing its `$id`→schema registry, plus a barrel export and CLI wiring.

## What's here
- **`TypeScriptGenerator` + `npm run schema:gen-types`** — emits one named TypeScript type per OCF schema:
  - enums → string-literal unions, scalars → aliases, objects → interfaces with `allOf`-inherited properties flattened, `$ref`s → named types, `oneOf`/`anyOf` → unions, `const` → literals.
  - Backwards-compat wrappers (`PlanSecurity*`) alias their parent with the `object_type` discriminant pinned, so they inherit every field instead of collapsing to a near-empty interface.
  - Validation facets (`format`, `pattern`, `min*`/`max*`, `uniqueItems`, `deprecated`) surface as JSDoc tags.
  - On the one base-name collision, the earliest `/schema` partition keeps the clean name (`types/File` → `OCFFile`; abstract `primitives/files/File` → `OCFFilesFile`).
- **`SchemaDereferencer`** — recursively inlines every `$ref` into a fully self-contained schema; wired into `ComposeSchemas` via `--dereference`.
- **`index.ts`** — public exports for the toolkit.

## Notes
- Generated output (`build/`) is produced by the scripts and is gitignored.
- All 68 tests pass; the generated `ocf.ts` module type-checks clean under `tsc --strict`.


MattCantor (approved):
This is great.  Two things come to mind:

### 1. Aggregate discriminated-union types

I think we should consider emitting aggregate types on top of the per-schema types:

- `AnyObject` - union of all concrete objects
- `AnyTransaction` - union of just the transactions
- `AnyFile` - union of the file wrappers
- `ObjectTypeMap` - mapping `object_type` to concrete type
- `FileTypeMap` - mapping `file_type` to concrete type

The OCF-Tools roadmap expects we'll have these in place:
- **Phase 2 ("type the machine context/events")** needs the transaction union as the XState event type. The machine gets a stream of transactions, and the event type is a discriminated union on `object_type`, i.e., `AnyTransaction`.
- **Phase 3 ("collapse the handlers into a data-driven table")** needs `ObjectTypeMap` so a table keyed by `object_type` hands each handler the correctly-typed transaction. And because the table is keyed on `keyof ObjectTypeMap`, it is exhaustive at compile time. This complements the planned "*" wildcard in the roadmap, which would catch unknown types at runtime.

Without these, OCF-Tools would need to rebuild the unions and the type lookup by hand. And the generator already computes the discriminant -> type mapping, so we could do it here as the single source of truth.

If we do this then:
- We'll need to include the backwards-compatibility alias values in the maps
- We could also have the generator throw if two schemas claim the same `object_type`/`file_type` value.

### 2. Testing the generated output

This PR regenerates the types to a temp file and confirms that tsc emits no errors. This confirms that the output compiles, but we could also test that every schema actually produced a type by asserting that there is only one export per emitted schema, plus the aggregates. Should work fine with the generated file gitignored.

~Happy to open up a follow-on PR for the aggregates and the test if that's better than folding them into this PR~

A proposed implementation of the above is in #585

PR #584, opened 2026-06-25
VersionWrapper: versioned schema dispatcher pattern + x-ocf-stability flags

### LINKED ISSUE #582: Versioned schema dispatcher pattern + schema stability flags
We want a structured way to evolve an object's data shape across a breaking change without forcing every consumer to cut over at once. Right now the closest thing we have is the Plan Security / Equity Compensation compatibility wrapper, but that's an aliasing trick (one shape, multiple object_type names). It doesn't help when the actual data shape needs to change.

Proposing a "versioned dispatcher" pattern:

- The stable, public `$id` moves onto a thin wrapper schema whose body is an `anyOf` of `$ref`s to concrete versioned shapes.
- Each versioned shape lives in its own file with a version suffix (`EquityCompensationIssuance.v1.schema.json`, `.v2.schema.json`) and its own `$id`. v1 is the current shape, unchanged.
- Consumers that already reference the public `$id` change nothing and transparently accept both shapes during the transition.
- Cutover at a major version is a one-line edit: drop the v1 ref from the dispatcher's `anyOf` and remove the v1 file.

Example dispatcher:

```json
{
  "$id": ".../issuance/EquityCompensationIssuance.schema.json",
  "anyOf": [
    { "$ref": ".../issuance/versions/EquityCompensationIssuance.v1.schema.json" },
    { "$ref": ".../issuance/versions/EquityCompensationIssuance.v2.schema.json" }
  ]
}
```

We also want a way to flag a schema (or a specific version shape) as pre-release or alpha:

- Add a stability marker to schemas, for example an `$comment` convention or a custom keyword like `x-ocf-stability: stable | beta | alpha | deprecated`.
- A new version shape can ship as `alpha` while the prior version stays `stable`, so the dispatcher can advertise the upcoming shape without anyone treating it as final.
- Tooling and consumers can then opt in or out of pre-release shapes (e.g. strict consumers validate against stable-only, early adopters include alpha).
- This also gives us a cleaner story than overloading version numbers, and it pairs with the dispatcher: stability lives on the version files, the dispatcher just lists them.

Things to work through on the OCF side:

1. Adopt the `.v#` filename convention and a `versions/` subfolder layout (or whatever placement we prefer). Build/validation tooling needs to recognize it.
2. Decide on the stability flag mechanism (custom keyword vs `$comment` convention) and the allowed values, and define what each level means for consumers.
3. draft-07 `additionalProperties: false` does not see properties pulled in through `$ref`/`anyOf`, so each versioned shape has to stay fully self-contained (its own complete properties, required, and `additionalProperties: false`). The dispatcher must not set `additionalProperties: false`.
4. Use `anyOf` rather than `oneOf`, since a document will often satisfy both shapes during the overlap window.
5. Tooling that assumes flat, property-bearing schemas (mapping generators, docs) needs to follow `anyOf`/`$ref` down to the versioned shapes, or operate on the version files directly.

If this looks reasonable I can prototype it against one transaction family (issuance) as a reference and template it across the rest.

### LINKED ISSUE #583: Doc generator: render versioned subschemas into one parent doc with stability flags
Companion to #582 (versioned schema dispatcher pattern). Once an object's public `$id` is a dispatcher that `anyOf`-refs versioned shapes, the doc generator needs to understand that structure instead of treating each file as its own flat schema.

What we want:

- For a dispatcher schema, produce a single markdown doc at the parent (public `$id`) level that wraps all of its versioned subschemas, rather than separate disconnected pages per version file.
- Within that parent doc, render each version shape as its own section and flag its stability: supported/stable vs deprecated vs alpha (matching whatever stability marker we land on in #582).
- Make the deprecated and alpha shapes visually distinct so readers can tell at a glance which shape is the current recommended one, which is on the way out, and which is not final yet.

Things to work through:

1. Detect dispatcher schemas (presence of `anyOf` over `$ref`s, or whatever signal we standardize on) and resolve the refs to pull in each version shape.
2. Read the stability flag off each version shape and surface it in the rendered output (section heading, badge, or callout).
3. Decide ordering and defaults (e.g. stable first, then alpha, then deprecated), and whether deprecated/alpha sections are collapsed by default.
4. Confirm anchors/links stay stable so existing links to the parent doc don't break.

Depends on the dispatcher and stability-flag decisions in #582.

JSCRUDAT0:
## Summary

Templates a **VersionWrapper** (versioned-schema dispatcher) pattern alongside the existing Compatibility Wrapper, plus a structured schema-stability flag. Closes #582 and #583.

A dispatcher is a thin schema whose stable, public `$id` resolves (via `anyOf`) to one of several self-contained, versioned shapes (`.v#` files in a `versions/` subfolder). Consumers that reference the public `$id` transparently accept any active version during a transition window; cutover at a major version is a one-line edit.

**No real schema adopts the pattern yet** — this lands the convention and the tooling, exercised entirely by fixtures/tests, as requested.

## Stability flag (#582)

A first-class JSON-Schema annotation keyword — **not** a `$comment` convention:

| `x-ocf-stability` | Meaning |
| --- | --- |
| `stable` | Current recommended shape (the default when absent) |
| `beta` | Feature-complete, still subject to change |
| `alpha` | Pre-release; not final, may change or be withdrawn |
| `deprecated` | On its way out; removal at a future major version |

## Changes

- **`SchemaComposer`** — `isVersionWrapper`, `versionRefsOf`, and stability helpers (`stabilityOf`, `STABILITY_LEVELS`, `STABILITY_RANK`), mirroring `isBackwardsCompatibleWrapper`. Dispatchers are left un-flattened so their `anyOf` body is preserved.
- **Doc generator (#583)** — `VersionedObjectSchemaNode` renders **one** page at the parent `$id`, folding each version in as its own section with a stability badge, ordered stable → beta → alpha → deprecated (alpha/deprecated visually distinct). Versioned shapes (`VersionedSubschemaNode`) get no standalone page; their property/source links are computed relative to the parent page so existing links stay stable.
- **Validator (#582)** — extracted a unit-testable `buildObjectTypeSchemaMap` that maps every versioned shape's `object_type` to the dispatcher's public `$id` (so an item validates via the `anyOf` and accepts any version); `validate.mjs` registers the `x-ocf-stability` keyword so AJV strict mode accepts it.
- **Docs** — `DesignPatterns.md` documents the dispatcher, the draft-07 self-containment rules (each shape fully self-contained with its own `additionalProperties: false`; the dispatcher sets none), and the stability levels.

## Detection

Structural and unambiguous: a dispatcher is an `anyOf` of bare `$ref`s with no own `properties`. Verified that **no existing schema** matches this signature, so nothing is misclassified (objects that use a top-level `anyOf` for conditional constraints carry `properties` and non-`$ref` branches).

## Testing

- **75 tests pass** (was 41) across new/extended suites: composer, doc generator, the `object_type` map builder, and an end-to-end AJV test proving a dispatcher accepts multiple version shapes and rejects non-matching instances.
- All three `validate.mjs` scripts and `docs:generate` run cleanly on the real schemas — generated doc output is byte-identical, so existing behavior is untouched.
- `prettier` clean.

Closes #582
Closes #583

### DIFF
--- a/docs/explainers/DesignPatterns.md
+++ b/docs/explainers/DesignPatterns.md
@@ -126,3 +126,75 @@ undocumented fields and types, which could cause:
 
 If you need to store additional metadata, use the `comments` field available on all OCF Objects, or
 maintain a separate mapping table in your system that links OCF object IDs to your custom data.
+
+## Evolving an Object's Shape: the VersionWrapper Dispatcher
+
+OCF already has a **Compatibility Wrapper** (see, e.g.,
+[PlanSecurityIssuance](../schema_markdown/schema/objects/transactions/issuance/PlanSecurityIssuance.md)),
+but that pattern is an _aliasing_ trick — one data shape exposed under multiple `object_type` names.
+It does not help when the underlying **data shape itself** needs to change across a breaking
+version.
+
+The **VersionWrapper** (a "versioned dispatcher") is the shape-changing analogue:
+
+-   The stable, public `$id` lives on a thin **dispatcher** schema whose body is an `anyOf` of
+    `$ref`s to concrete, versioned shapes — and nothing else (no `properties` of its own, no
+    `additionalProperties`).
+-   Each versioned shape lives in its own file under a `versions/` subfolder, following the `.v#`
+    filename convention (`EquityCompensationIssuance.v1.schema.json`,
+    `EquityCompensationIssuance.v2.schema.json`), with its own `$id`. `v1` is the current shape,
+    unchanged.
+-   Consumers that already reference the public `$id` change nothing and transparently accept every
+    listed shape during the transition window.
+-   Cutover at a major version is a one-line edit: drop the retiring `$ref` from the dispatcher's
+    `anyOf` and remove the stale version file.
+
+```json
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": ".../issuance/EquityCompensationIssuance.schema.json",
+    "anyOf": [
+        { "$ref": ".../issuance/versions/EquityCompensationIssuance.v1.schema.json" },
+        { "$ref": ".../issuance/versions/EquityCompensationIssuance.v2.schema.json" }
+    ]
+}
+```
+
+A few rules make this work with draft-07:
+
+1.  Use `anyOf`, **not** `oneOf` — during the overlap window a document will often satisfy more than
+    one shape, and `oneOf` would reject it for matching more than one branch.
+2.  draft-07's `additionalProperties: false` does not see properties pulled in through
+    `$ref`/`allOf`, so each versioned shape must be **fully self-contained** (its own complete
+    `properties`, `required`, and `additionalProperties: false`). The dispatcher itself must **not**
+    set `additionalProperties`.
+3.  Tooling that walks schemas (the validator, the doc generator) follows the dispatcher's `anyOf`
+    down to the versioned shapes. The validator maps every versioned shape's `object_type` to the
+    dispatcher's public `$id`, so validating an item against that `object_type` accepts any active
+    version.
+
+## Stability Flags (`x-ocf-stability`)
+
+A schema — or a single versioned shape — can be flagged as pre-release or on its way out with the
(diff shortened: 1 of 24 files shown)


JSv4 (approved):
Adversarial LLM reviews X2 - OK

PR #586, opened 2026-06-29
Add --experimental flag to resolve version dispatchers (none/compatibility/unstable)

JSCRUDAT0:
## What

Adds an `--experimental` flag to the three tools that consume the OCF schema — `schema:compose`, `schema:gen-types`, and `docs:generate` — to choose how a [VersionWrapper dispatcher](docs/explainers/DesignPatterns.md) (`#584`) is resolved. All three accept the same flag and resolve dispatchers identically, so the composed schemas, generated types, and docs always agree.

| `--experimental` | Behavior for a dispatcher |
| --- | --- |
| `compatibility` **(default)** | Keep the dispatcher intact: the public `$id` is the **union** of every listed shape (codegen emits `V1 \| V2 \| …`; docs fold all versions into one page). Unchanged from prior behavior. |
| `none` | Drop the dispatcher; expose only the **latest `x-ocf-stability=stable`** shape under the public `$id`. Pre-release / older shapes leave the public surface. (No stable shape → latest overall + warning.) |
| `unstable` | Expose the **latest `alpha`/`beta`** shape, falling back to the latest `stable` when none is pre-release. |

## How

A single shared pre-pass, `applyExperimentalMode(rawSchemas, mode)`, in `SchemaComposer`:

- `compatibility` → no-op pass-through; the existing dispatcher-aware code paths handle the union / folded page.
- `none` / `unstable` → collapse each dispatcher to its selected versioned shape (re-homed onto the dispatcher's stable public `$id`) and remove the other version shapes entirely. After this pass there are **no dispatchers left**, so the composer, codegen, doc generator, and dereferencer all operate on ordinary schemas with zero per-tool dispatcher special-casing.

"Latest" is the highest `.v#` number among the candidate shapes. A property that references a dispatcher's public `$id` keeps working — that `$id` now resolves to the collapsed shape.

Also: stop writing vacuous `"properties": {}` / `"required": []` to composed/dereferenced **enum and scalar** output files (`omitEmptyComposedContainers`).

## Testing

- 175 unit tests pass (23 suites), prettier clean. New coverage for version selection, `applyExperimentalMode` collapse/drop, TS-gen `none`/`unstable`, doc-gen collapse, and the output cleanup.
- Exercised end-to-end on a real dispatcher fixture through the **compose** CLI (all 3 modes incl. `--dereference`: correct collapse, version files dropped, public `$id` preserved) and the **docs** generator (`compatibility` = folded page with v1+v2 + STABLE/ALPHA badges; `none` = plain page with the stable shape; `unstable` = plain page with the alpha shape).

## Notes

- Default is `compatibility`, so existing output is unchanged until a consumer opts into `none` / `unstable`.
- Branches off `main` (which now includes #584). The diff is exactly the `--experimental` feature.

### DIFF
--- a/docs/explainers/DesignPatterns.md
+++ b/docs/explainers/DesignPatterns.md
@@ -226,3 +226,32 @@ and ordered `stable` → `beta` → `alpha` → `deprecated`, so a reader can te
 is current, which is not final, and which is on the way out. This works for object, type, and enum
 dispatchers alike — each version section renders the body appropriate to its kind (an object/type
 property table, or an enum's value list).
+
+## Choosing Which Versions Ship: the `--experimental` Flag
+
+The dispatcher carries every active versioned shape, but most consumers only want **one**. The
+`--experimental` flag tells the toolchain — the schema composer (`schema:compose`), the TypeScript
+codegen (`schema:gen-types`), and the documentation generator (`docs:generate`) — how to resolve a
+dispatcher. All three accept the same flag and resolve dispatchers identically, so the composed
+schemas, the generated types, and the docs always agree.
+
+| `--experimental` | What it emits for a dispatcher                                                                                                                                                                                                                                       |
+| ---------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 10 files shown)


JSv4 (approved):
adversarially reviewed

PR #587, opened 2026-06-29
Add experimental vesting schemas behind version dispatchers; mark vesting-start planned_deprecation

JSCRUDAT0:
## What & why

Wires the new canonical vesting model in as the **experimental** version of the existing vesting schemas, using the version-dispatcher / `--experimental` feature so that **both** surfaces generate from one `/schema` tree:

- **stable** (`--experimental=none`) → today's vesting schemas
- **experimental** (`--experimental=unstable`) → the new model, minus the removed start event
- **compatibility** (default) → unchanged from `main` (dispatchers stay `V1 | V2` unions)

## Changes

**Version dispatchers** (public `$id` unchanged; `v1` = current `stable` shape, `v2` = new `alpha` shape, both keeping the same `object_type` so routing/aggregates are stable):

| Object | v1 (stable) | v2 (alpha) |
|---|---|---|
| `EquityCompensationIssuance` | `vesting_terms_id` | `vesting_template_id` + per-grant `vesting_start_date` |
| `VestingEvent` | `vesting_condition_id` | named `event_id` |
| `VestingTerms` | condition DAG (`vesting_conditions`) | ordered statement template (`statements`) |

New self-contained draft-07 types: `VestingStatement`, `VestingScheduleSegment`, `VestingScheduleCliff`, `VestingEventCondition` (the canonical `schedule`/`cliff`/`event_condition` shapes, extracted into named `$ref` types so the docs renderer accepts them).

**`planned_deprecation` (a new stability level — *not* `deprecated`):** the vesting-commencement-start transaction (`TX_VESTING_START`) was removed in the new model, with its role absorbed by `EquityCompensationIssuance.vesting_start_date`. It becomes a dispatcher whose single shape carries `x-ocf-stability: planned_deprecation`:
- **kept and fully active** under `none` / `compatibility`
- **dropped entirely** under `unstable` — `applyExperimentalMode` now drops a planned-deprecation-only dispatcher and prunes the resulting dangling `$ref` (in `TransactionsFile`) and the stale `TX_VESTING_START` value in the `ObjectType` enum, so the experimental surface still composes / codegens / documents cleanly.

## Verification

| | composed | TS types (AnyObject / AnyTx) | docs |
|---|---|---|---|
| **stable** (`none`) | 176 — VestingStart present | 49 / 40 | ✓ |
| **experimental** (`unstable`) | 175 — **VestingStart absent**, no `TX_VESTING_START` anywhere | 48 / 39 | ✓ no VestingStart page |
| **compatibility** (default) | 183 | **49 / 40** (matches the integration-test tripwire) | ✓ dispatcher pages |

- `npm test` — **192/192** (incl. 9 new tests for the stability level, the `unstable` drop, and ref/enum pruning)
- `schema:validate-ocf-file-schemas`, `schema:validate-example-ocf-files`, `schema:validate-all-objects-have-samples` — all pass
- `tsc --noEmit` clean; prettier clean
- Regenerated the markdown docs (`docs/schema_markdown/...`)


JSCRUDAT0:
@arthur-clara and @MattCantor, can you guys give me your feedback on this approach to tracking planned deprecation and allowing optionally planned shapes before they're approved / officially added? This will let us move the vesting schemas Matt has been working on into OCF repo proper (where they belong) without technically changing schema spec.

JSCRUDAT0:
@MattCantor, did I get your new schemas right?

MattCantor:
> @MattCantor, did I get your new schemas right?

I'm looking through this. First thing that jumps out is `VestingDayOfMonth.schema.json`. I don't think we're going to abandon the concept entirely. Instead I think we'll need to decide which vesting day of month policies survive and which are abandoned.

Per [this issue in OCF-Composed-Schemas](https://github.com/Open-Cap-Table-Coalition/OCF-Composed-Schemas/issues/131), there is some minimal overlap between Carta's vesting day of month policies and OCF's vesting day of month policies. Probably need to decide which survive. But the we'll continue to need the concept going forward.

UPDATE: now I see why you concluded that vesting day of month was going to be deprecated. I never plumbed it into the new v2 shapes!

EDITED: I referenced allocation types in my original comment, I mean vesting day of month policies

JSCRUDAT0:
> > @MattCantor, did I get your new schemas right?
> 
> I'm looking through this. First thing that jumps out is `AllocationType.schema.json`. I don't think we're going to abandon the concept entirely. Instead I think we'll need to decide which allocation types survive and which are abandoned.
> 
> Per [this issue in OCF-Composed-Schemas](https://github.com/Open-Cap-Table-Coalition/OCF-Composed-Schemas/issues/131), there is some minimal overlap between Carta's allocation types and OCF's allocation types. Probably need to decide which survive. But the we'll continue to need the concept going forward.
> 
> UPDATE: now I see why you concluded that allocation type was going to be deprecated. I never plumbed it into the new v2 shapes!

Yup, I marked unused ones to pending_deprecation. The good thing about the versioning system though is the official schemas wouldn't change once composed. The new subschema versioning lets us mark schemas and properties for future deprecation. You can set a flag when you 1) compose schemas, 2) generate types, or 3) generate docs. In compatibility mode, you get beta schemas, plus stable schemas plus pending deprecation schemas. In unstable mode, you get latest schemas, whether beta or stable, minus deprecated schemas. In stable mode, you get current version of the schemas. 

I'd really like to take a stab at what vesting looks like and keep it in the OCF repo proper. It was getting really challenging to think about how to keep things in sync across repos, with some of the schema in OCF-Composed-Schemas and some here. I'm planning to keep working in the OCF-Composed-Schema repo off of the unstable schemas generate by this PR / branch. Feel free to edit the vesting schemas here before we merge. Good news (Again) with versioning is we don't have to get the alpha schemas perfectly right to continue working on them. We can merge them in as alphas and graduate them later (at which point the changes become breaking and we need a version 2).

MattCantor:
This versioning approach is great. And yes, you read the new schemas right. Two confirmations and one fix-up, which I've pushed straight to the branch:

**1. Allocation (`CUMULATIVE_ROUND_DOWN`).** Correct as-is. v2 locks in `CUMULATIVE_ROUND_DOWN` as the only available allocation policy, so dropping `allocation_type` and flagging `AllocationType` as `planned_deprecation` is right.

**2. `vesting_start_date` required with a template.** The `dependencies: { vesting_template_id: ["vesting_start_date"] }` block correctly makes `vesting_start_date` required only when a template is referenced. This enables the far-future date placeholder convention for contingent starts. And it would be an ignored field when it is provided along with `vestings` but no `vesting_template_id`. This is fine, basically a no-op, or maybe later we could have the OCF-Tools validator fire a warning when they don't match.

**3. Vesting day-of-month.** This only looked like it was v1-only because I hadn't wired it into v2. The concept survives, but in this alpha version of v2 we're narrowing it:

- a new **`VestingDayOfMonthPolicy`** enum (`alpha`) is the v2 successor to the 32-value `VestingDayOfMonth`, narrowed to Carta's set: `VESTING_START_DAY` (default), `FIRST_DAY_OF_MONTH`, `LAST_DAY_OF_MONTH`, `VESTING_START_DAY_MINUS_ONE`.
- an optional `vesting_day_of_month` field on `VestingScheduleSegment` referencing it (default `VESTING_START_DAY`). This matches Carta, which carries `vestingOccurs` on each `VestingPeriod` (its term for a segment), so each segment can carry its own day-of-month policy. The per-grant `vesting_start_date` stays on the issuance, and each `VESTING_START_DAY` or `VESTING_START_DAY_MINUS_ONE` resolves against it.
- left the old `VestingDayOfMonth` (32 values, `planned_deprecation`) **untouched**.
- **Successor enum**, rather than narrowing the old one in place, because that would be a breaking change to the published v1 enum. I also don't think versioning the same enum name would work because an enum dispatcher's `compatibility` `oneOf` would accept the *union* of both value sets. The successor enum keeps every surface strict and leaves your `planned_deprecation` flag on the old enum correct.
- **Explicit `LAST_DAY_OF_MONTH`** dissolves the ambiguity around whether OCF's v1 `VESTING_START_DAY_OR_LAST_DAY_OF_MONTH` was supposed to infer end-of-month intent when used with a vesting start on day 29, 30, 31. With the v2 enum, `VESTING_START_DAY` clamps but does not infer end-of-month intent, and `LAST_DAY_OF_MONTH` should be used when end-of-month is intended.

MattCantor (approved):

MattCantor:
Follow-up: pushed a second change to the branch. Tightening the v2 `VestingStatement` typing.

**What changed.** Split `VestingStatement` into two named branch types — `ScheduledVestingStatement` (a `schedule`, optionally gated by an `event_condition`) and `MilestoneVestingStatement` (an `event_condition`, no schedule) — and expressed the choice as an inline `oneOf` in `VestingTerms.v2` `statements.items`. The standalone `VestingStatement` type is removed.

**Why.** The single all-optional object lowered to `{ order, percentage, schedule?, event_condition? }`, which drops the "at least one axis" constraint at the type level. The branch form lowers to a proper discriminated union: `statements: (OCFScheduledVestingStatement | OCFMilestoneVestingStatement)[]`. Pure refactor — the same documents validate and the neither-axis corner is still rejected.

**Convention.** I kept this to OCF's existing point-of-use union pattern: the union lives inline in the consuming property, pointing at standalone branch types — exactly like `VestingCondition.trigger`, `ConversionRight.conversion_mechanism`, and `TransactionsFile.items`. No standalone named union type (OCF has none today), so no doc-generator change. (An earlier attempt at a standalone `oneOf` *type* did break `docs:generate`. The generator has no node for a non-dispatcher type-level union. Which confirmed the inline form is the idiomatic one.)

**On `order`.** Both branches keep it. A scheduled statement chains implicitly by `order`; a milestone records its position but, having no `schedule`, does not affect the timing of other statements. This was confirmed by Ray.

